PR #9, opened 2022-04-05
Support for QuestionSet and Quiz models + creating Quiz

dalmia:
Schema: https://www.notion.so/avantifellows/Schema-65640e37928848bbaf0f0c001fe9da47
DB Design: https://www.notion.so/avantifellows/Database-4cfd0b2c9d6141fd88197649b0593318


app/models.py
```diff
@@ -69,3 +83,59 @@ class Config:
...
+    questions: List[Question]
+
+
+class Quiz(BaseModel):
+    id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
+    question_sets: List[QuestionSet]
+    maxMarks: int
+    numGradedQuestions: int
+    shuffle: bool = False
+    numAttemptsAllowed: int = 1
```
deepansh96, line 99: snake case in all places!

PR #13, opened 2022-04-14
Added .env

### LINKED ISSUE #3: Hide database credentials with environment variables

Salonichoudhary07:
Fixes #3 
Added environment files

### DIFF
--- a/app/database.py
+++ b/app/database.py
@@ -1,4 +1,4 @@
 from pymongo import MongoClient
 import os
 
-client = MongoClient(os.environ.get("MONGO_CLIENT"))
+client = MongoClient(os.environ.get("MONGO_CLIENT", ""))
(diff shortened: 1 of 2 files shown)


dalmia:
Screenshot 2022-04-18 at 10 21 01 AM
- where in the code is the environment variable being set?
- you need to update the installation document to say that one needs to copy `.env.example` to `.env` and fill in the values
- how would we be differentiating between the environment variables for the local set up, staging and production?
- you need to create a `docs/ENV.md` file to explain each environment variable being add in `.env.example`.

app/database.py
```diff
@@ -1,3 +1,4 @@
 from pymongo import MongoClient
-client = MongoClient(
-    "mongodb+srv://quiz:p$#p4h7y_Z44R-n@cluster0.uocfg.mongodb.net/quiz?retryWrites=true&w=majority")
+import os
+
+client = MongoClient(os.environ.get("MONGO_CLIENT"))
```
dalmia, line 4:
```suggestion
client = MongoClient(os.environ.get("MONGO_AUTH_CREDENTIALS"))
```

dalmia:
being completed here #26

PR #20, opened 2022-04-19
Adds endpoints for sessions and session_answers

### LINKED ISSUE #19: Add endpoints for sessions and session_answers

dalmia:
Fixes #19 

Schema for Sessions and Session Answers: https://www.notion.so/avantifellows/User-Quiz-Sessions-Schema-a19339ede2ee41dd98af40927fb02086
Details on the subset pattern and how we are using it for sessions and session answers: https://www.notion.so/avantifellows/Database-4cfd0b2c9d6141fd88197649b0593318

### Major
- support for creating new sessions for a given user-quiz pair. If this is the first session for the pair, new session answers are created with all answers as `None`. If a previous session already exists between the pair, the answers from that session are used to instantiate the answers of this session. We use the subset pattern here where the first few answers are stored in respective session document itself and all the session answers are stored separately in another collection. (the specific part of the subset pattern where only 10 or so session answers are stored and not the rest hasn't been done yet as we didn't do it for question/question-set too. Will do that in a separate PR. 
- added support for updating session answers. while updating it is important to update the instance of the session answer in the `session_answers` collection as well as in the `sessions` collection if that session answer is present in the subset of session answers linked to that session.
- created new models for each of Question, Quiz, Session, SessionAnswer for the response, different from the model used for the body. Read about Response Models in FastAPI's documentation [here](https://fastapi.tiangolo.com/tutorial/response-model/).

### Minor
- made attributes in QuizMetadata which haven't been implemented in the frontend yet optional
- changed the type of `correct_answer` from `Optional[List[int]]` to `Union[List[int], None]` because it is not that when the correct answer is specified, it is always going to be a list of integers. So, this one better reflects that and enables us to easily add more types when we support other types of correct answers.
- removed the API calls for creating a question directly and for listing all questions as we don't need them for now. We want questions to only be created via the `/quiz` endpoint.
- replaced status code numbers with variables from `fastapi.status` for better readability.


app/models.py
```diff
@@ -85,12 +87,46 @@ class Config:
         }
 
 
+class QuestionResponse(Question):
+    """Model for the response of any request that returns a question"""
+
+    question_set_id: str
```
deepansh96, line 93: when we want the backend to return a question, why are we just returning the question_set_id?

app/models.py
```diff
@@ -137,5 +193,147 @@ class Config:
                 ],
                 "max_marks": 10,
                 "num_graded_questions": 3,
+                "metadata": {"quiz_type": "JEE", "subject": "Maths", "grade": "8"},
+            }
+        }
+
+
+class QuizResponse(Quiz):
+    """Model for the response of any request that returns a quiz"""
```
deepansh96, line 202: we're not returning anything here?

app/models.py
```diff
@@ -137,5 +193,147 @@ class Config:
...
+            }
+        }
+
+
+class SessionAnswer(BaseModel):
+    """Model for the body of the request that creates a session answer"""
+
+    id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
+    question_id: str
+    answer: Union[List[int], None] = None
```
deepansh96, line 270: this should also have the session_id linked here right?

app/models.py
```diff
@@ -137,5 +193,147 @@ class Config:
...
+        schema_extra = {"example": {"question_id": "4567", "answer": [0, 1, 2]}}
+
+
+class UpdateSessionAnswer(BaseModel):
+    """Model for the body of the request that updates a session answer"""
+
+    answer: Union[List[int], None]
+
+    class Config:
+        schema_extra = {"example": {"answer": [0, 1, 2]}}
```
deepansh96, line 285: specifying only `answer` here means that no one will be able to update anything other than the answer?
Like I can't update the session_id or user_id if I wanted to?

app/routers/questions.py
```diff
@@ -1,29 +1,16 @@
...
-@router.post("/", response_model=Question)
-async def create_question(question: Question):
-    question = jsonable_encoder(question)
-    new_question = client.quiz.questions.insert_one(question)
-    created_question = client.quiz.questions.find_one({"_id": new_question.inserted_id})
-    return JSONResponse(status_code=status.HTTP_201_CREATED, content=created_question)
-
-
-@router.get("/{question_id}")
+@router.get("/{question_id}", response_model=QuestionResponse)
```
deepansh96, line 8: because of the response model, when we make a get request to get a question with some `question_id`, it will not return the question ka details but just the question_set_id right?
Why is that? I just want a question so I should be able to get it?

Will we have to make another call to get question from the question_set_id?

app/routers/sessions.py
```diff
@@ -0,0 +1,81 @@
...
+                    )
+    else:
+        session["is_first"] = False
+
+        # restore the answers from the previous sessions
+        session_answers = list(
+            client.quiz.session_answers.find({"session_id": previous_session["_id"]})
+        )
+        if session_answers:
+            for index, session_answer in enumerate(session_answers):
```
deepansh96, line 53: No need of this if statement. The enumeration won't happen if session_answers is none or empty

app/routers/sessions.py
```diff
@@ -0,0 +1,81 @@
...
+                for key in ["_id", "session_id"]:
+                    session_answer.pop(key)
```
deepansh96, line 55: why are we removing data from session_answer?

PR #24, opened 2022-04-22
Fixes the ordering of the documents in various collections

### LINKED ISSUE #23: Fix order in which documents of different collections are returned

dalmia:
Fixes #23 

## Summary
- the default ordering of the objects when queried is not in ascending or descending order of their creation 
- we need a fixed order when we fetch documents from various collections. e.g. during the creation of a new session, I want to search only for the immediately previous session - sorting matters here.


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,18 @@
+<!--
+  Thanks for submitting a pull request!
+  We appreciate you spending the time to work on these changes. Please provide enough information so that others can review your pull request.
+
+  Before submitting a pull request, please make sure the following is done:
+
+  1. Fork [the repository](https://github.com/avantifellows/quiz-backend) and create your branch from `master`.
+  2. Run the installation steps from the project's [README.md](https://github.com/avantifellows/quiz-backend#readme).
+  3. Please ensure coding standard and conventions are followed. You can find the details at https://vuejs.org/v2/style-guide/#Priority-A-Rules-Essential-Error-Prevention.
```
deepansh96, line 9: The backend is not in vue right?

app/routers/sessions.py
```diff
@@ -29,7 +31,9 @@ async def create_session(session: Session):
         # since we know that there is going to be only one question set for now
         if "question_sets" in quiz and quiz["question_sets"]:
             question_set_id = quiz["question_sets"][0]["_id"]
-            questions = client.quiz.questions.find({"question_set_id": question_set_id})
+            questions = client.quiz.questions.find(
+                {"question_set_id": question_set_id}, sort=[("_id", pymongo.ASCENDING)]
+            )
```
deepansh96, line 36: why do we need questions in ascending order?

pritamps: Why sort by `id`? If we switch to a random UUID (isn't it already random), this will cause a problem, no? Questions should have an ordering field in the quiz, I think

app/routers/sessions.py
```diff
@@ -47,7 +51,10 @@ async def create_session(session: Session):
 
         # restore the answers from the previous sessions
         session_answers = list(
-            client.quiz.session_answers.find({"session_id": previous_session["_id"]})
+            client.quiz.session_answers.find(
+                {"session_id": previous_session["_id"]},
+                sort=[("_id", pymongo.ASCENDING)],
```
pritamps, line 56: Why sort again? We will match question ids to show previously answered questions, no?

.github/PULL_REQUEST_TEMPLATE.md
dalmia: I remember linking PEP8 ka style guide here. No idea why that is not the case :/

app/routers/sessions.py
dalmia: in the session object, we are storing the first 10 session answers right? (the subset pattern)
so, the session answers have to be stored in the order of the questions in the quiz. If I don't sort it, the questions appear in some random order.

@pritamps the ID generated is already a UUID - only the name of the field is ID. Example image added!
Screenshot 2022-04-27 at 10 30 04 AM

app/routers/sessions.py
dalmia: please read my reply to the comment above - it is because we are storing the first few session answers in the session object itself, in which case, the session answer objects have to be stored in a fixed order.

deepansh96 (approved):

app/routers/sessions.py
pritamps: I think i have some basic misunderstanding of what's going on. Then how does sorting by ID work if it's supposed to be random?

pritamps (approved):
Just a quick question. You can merge, but answer it before merging :)

app/routers/sessions.py
dalmia: In my answer above, I didn't mention that the ID was random right? It is auto-incremented, but it is unique enough to do the job of the UUID. The random part that I did mention above was the randomness in the order of the questions returned if we do not sort!

dalmia:
> Just a quick question. You can merge, but answer it before merging :)

Done!

app/routers/sessions.py
pritamps: Aiyo... The purpose of using UUID is lost if you auto-increment them. A user with access to one UUID will then have access to all subsequent UUIDs in the session. So in this case, there's no real advantage to using UUID at all :-/

PR #27, opened 2022-05-04
Added test and coverage report in CI

### LINKED ISSUE #28: Add test and coverage report in CI
Make the tests run when a PR is created and also calculate the coverage for the same so that we can get a report of how much of the code is covered by tests,

Salonichoudhary07:
Fixes #28 

## Summary

Runs tests in CI and gives the coverage report.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,11 +36,11 @@ jobs:
         env:
           MONGO_AUTH_CREDENTIALS: "mongodb+srv://quiz:p$#p4h7y_Z44R-n@cluster0.uocfg.mongodb.net/quiz?retryWrites=true&w=majority"
         # command to run tests and generate coverage metrics
-        run: |
-          coverage run -m pytest
-          coverage report
-          coverage xml
+        script:
+          - coverage run -m pytest
+          - coverage report
+          - coverage xml
         coverage: '/TOTAL.*\s([.\d]+)%/'
         artifacts:
-        reports:
-          cobertura: coverage.xml
+          reports:
+            cobertura: coverage.xml
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,11 +31,10 @@ jobs:
           python -m pip install --upgrade pip
           pip install -r app/requirements.txt
           pip install pytest pytest-cov
-          install_compass
 
       - name: Run Test Cases
         env:
-          MONGO_AUTH_CREDENTIALS: "mongodb://mongo:27017/quiz"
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -13,3 +13,32 @@ jobs:
...
+      - name: Set up Python 3.9
+        uses: actions/setup-python@v2
+        with:
+          python-version: 3.9
+
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r app/requirements.txt
+          pip install pytest pytest-cov
```
dalmia, line 33: pytest is already in the requirements right?

.github/workflows/ci.yml
```diff
@@ -13,3 +13,32 @@ jobs:
...
       - uses: actions/setup-python@v2
       - uses: pre-commit/action@v2.0.3
+
+  test:
+    name: Test cases
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+
+      # Install Python.
```
dalmia, line 23:
```suggestion
      # Install Python
```

.github/workflows/ci.yml
```diff
@@ -13,3 +13,32 @@ jobs:
...
+
+      - name: Install Dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r app/requirements.txt
+          pip install pytest pytest-cov
+
+      - name: Run Test Cases
+        env:
+          MONGO_AUTH_CREDENTIALS: "mongodb+srv://quiz:p$#p4h7y_Z44R-n@cluster0.uocfg.mongodb.net/quiz?retryWrites=true&w=majority"
```
dalmia, line 37: why do we need this if we are running mongo mock?

app/requirements.txt
```diff
@@ -8,3 +8,4 @@ python-dotenv==0.20.0
 mongoengine==0.24.1
 pytest==7.1.2
 mongomock==4.0.0
+requests==2.27.1
```
dalmia, line 11: why do we need this?

dalmia (changes requested):

app/requirements.txt
Salonichoudhary07: I had installed it locally but had forgotten to add this.

.github/workflows/ci.yml
Salonichoudhary07: yes

.github/workflows/ci.yml
Salonichoudhary07: We will need this or else the post and get request will not be accepted, the mock needs to know which DB to mock.

PR #25, opened 2022-04-26
Added testing 

Salonichoudhary07:
Added testing for quiz backend.

PR #37, opened 2022-05-10
Adds support for tracking question visit status

### LINKED ISSUE #35: Add field to track if question has been visited

dalmia:
Fixes #35 

## Summary
- new field `visited` added to `SessionAnswer` to track if that question has been visited by the user or not
- updated the API call to allow updating the `visited` field.
- modularized the code for the tests

Explanation for the method defined in `utils.py`:

We have kept the model `UpdateSessionAnswer` for specifying everything that can be passed to the API call for updating a session answer. I have listed both the fields as optional so that one could specify only either of them or both.

Now, when I am updating the `answer` part of a session answer, I am not passing the `visited` field and I want the value of that field to be unaffected. However, by default, `pydantic` still includes `visited` in the model instance that is used to update the corresponding session answer document and sets its value to null. The method has been added to prevent that from happening. Look more [here](https://stackoverflow.com/questions/65362524/in-json-created-from-a-pydantic-basemodel-exclude-optional-if-not-set).


app/tests/base.py
```diff
@@ -0,0 +1,38 @@
...
+    def setUpClass(cls):
+        connect("mongoenginetest", host="mongomock://127.0.0.1:8000")
+        cls.client = TestClient(app)
+
+    @classmethod
+    def tearDownClass(cls):
+        disconnect()
+
+    def setUp(self):
+        data = open("app/dummy_data/homework_quiz.json")
```
deepansh96, line 21: Also this file is not committed to this branch?

app/tests/test_session_answers.py
```diff
@@ -0,0 +1,47 @@
...
+    def setUp(self):
+        super().setUp()
+        self.session_answers = self.session["session_answers"]
+        self.session_answer = self.session_answers[0]
+        self.session_answer_id = self.session_answer["_id"]
+
+    def test_gets_session_answer_with_valid_id(self):
+        response = self.client.get(f"/session_answers/{self.session_answer_id}")
+        assert response.status_code == 200
+        session_answer = json.loads(response.content)
```
deepansh96, line 15: ye content kahaan se aya?

deepansh96 (changes requested):

app/utils.py
```diff
@@ -0,0 +1,27 @@
...
+        if isinstance(value, dict):
+            node = destination.setdefault(key, {})
+            union(value, node)
+        else:
+            destination[key] = value
+
+    return destination
+
+
+def remove_optional_unset_args(model: BaseModel):
```
deepansh96, line 15: add that stackovrflow link as a comment please

and ye union waala function ko andar hi likh de na, inside `remove_optional_unset_args`?
For now that is going to be used here only.

If we need it kabhi we can take it out.
It will just be easier to understand for someone who sees the code after an year

deepansh96 (changes requested):

app/tests/base.py
```diff
@@ -0,0 +1,38 @@
...
+from fastapi.testclient import TestClient
+from mongoengine import connect, disconnect
+from main import app
+
+# client = TestClient(app)
+
+
+class BaseTestCase(unittest.TestCase):
+    @classmethod
+    def setUpClass(cls):
```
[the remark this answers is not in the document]
dalmia: So, whenever we define a classmethod, just like we use `self` usually, `cls` is the python convention for this!https://www.geeksforgeeks.org/classmethod-in-python/

app/tests/base.py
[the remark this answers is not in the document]
dalmia: agreed.

dalmia: This file was added beforehand in the PR where tests were first added: https://github.com/avantifellows/quiz-backend/pull/25/files

deepansh96 (approved):

PR #40, opened 2022-05-11
Added test for sessions

Salonichoudhary07:
Added tests for sessions

### DIFF
--- a/app/tests/test_sessions.py
+++ b/app/tests/test_sessions.py
@@ -1,0 +1,20 @@
+from .base import SessionsBaseTestCase
+
+
+class SessionsTestCase(SessionsBaseTestCase):
+    def setUp(self):
+        super().setUp()
+        self.session_id = self.session["_id"]
+
+    def test_gets_session_answer_with_valid_id(self):
+        response = self.client.get(f"/sessions/{self.session_id}")
+        assert response.status_code == 200
+        session = response.json()
+        for key in ["quiz_id", "user_id"]:
+            assert session[key] == self.session[key]
+
+    def test_get_question_returns_error_if_id_invalid(self):
+        response = self.client.get("/sessions/00")
+        assert response.status_code == 404
+        response = response.json()
+        assert response["detail"] == "session 00 not found"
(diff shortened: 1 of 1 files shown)


app/tests/test_sessions.py
```diff
@@ -0,0 +1,65 @@
+import json
+from .base import SessionsBaseTestCase
+
+
+class SessionsTestCase(SessionsBaseTestCase):
+    def setUp(self):
+        super().setUp()
+        self.session_id = self.session["_id"]
+
+    def test_gets_session_answer_with_valid_id(self):
```
dalmia, line 10: why `answer`?

app/tests/test_sessions.py
```diff
@@ -0,0 +1,65 @@
...
+        response = self.client.get(f"/sessions/{self.session_id}")
+        assert response.status_code == 200
+        session = response.json()
+        for key in ["quiz_id", "user_id"]:
+            assert session[key] == self.session[key]
+
+    def test_get_question_returns_error_if_id_invalid(self):
```
dalmia, line 17: `question`?

app/tests/test_sessions.py
```diff
@@ -0,0 +1,65 @@
...
+        response = self.client.post(
+            "/sessions/", json={"quiz_id": quiz["_id"], "user_id": 1}
+        )
+        assert response.status_code == 201
+        session = json.loads(response.content)
+        assert session["is_first"] is False
+        assert session["has_quiz_ended"] is False
+        assert len(session["session_answers"]) == len(
+            quiz_data["question_sets"][0]["questions"]
+        )
```
dalmia, line 65: keep this a separate test and for this part, you need to ensure that the session answers from the past session get copied over instead of being null. For that test, you can re-use the session that is already created in base and before you make this post call, you will need to update the session answers for the session created in base to make sure one or more of them have non-null answers which are being copied over.

app/tests/test_sessions.py
Salonichoudhary07: I actually corrected everything, maybe it all got undone. Sorry,

PR #42, opened 2022-05-16
Added Organization model and API

### LINKED ISSUE #41: Organization Model with API key support

sudheshna-donthineni:
Fixes #41 

## Summary

This code adds an Organization model to our database. 
- Contains an org name and org key as fields. 

Linked frontend PR: https://github.com/avantifellows/quiz-frontend/pull/52
For testing:
- Have created a dummy session.
http://localhost:8080/?sessionId=test-quiz
https://staging-auth.avantifellows.org/?sessionId=test-quiz


app/models.py
```diff
@@ -6,6 +6,10 @@
 answerType = Union[List[int], str, None]
 
 
+class Organization(BaseModel):
+    org_name: str
```
deepansh96, line 10: also add an `id` no?

app/routers/organizations.py
```diff
@@ -0,0 +1,19 @@
+from fastapi import APIRouter, status, HTTPException
+from database import client
+from models import Organization
+
+router = APIRouter()
+
+
+@router.get("/check-auth-token/{api_key}", response_model=Organization)
+async def get_auth_token(api_key: str):
```
deepansh96, line 9: This is not really "get" auth token right?
Maybe something like "verify_auth_token"?

app/routers/organizations.py
```diff
@@ -0,0 +1,19 @@
...
+
+    if (
+        org := client.quiz.organization.find_one({"org_key": api_key}, {"org_name": 1})
```
deepansh96, line 12: why `{"org_name": 1}`?
Why is there a need to sort by org name?

sudheshna-donthineni: It's projection, not sorting. 
Instead of returning all the fields in the query, it returns only the mentioned field. 
In this case, rather than returning the id, name and api_key, it returns only the org name.

app/models.py
sudheshna-donthineni: This is Response Model, no. We don't need id in our response, only org_name.

sudheshna-donthineni: Anyways, added in the next PR.

app/routers/organizations.py
deepansh96: understood
it was not super clear, maybe because of the syntax

I had two questions
- you're finding things here by `org_key` but I don't see any `org_key` defined in the model?
- why keep two separate names? `org_key` and `api_key` when they mean the same thing?

deepansh96 (approved):

PR #43, opened 2022-05-16
Added numerical type question

Salonichoudhary07:
## Summary
Added numerical type question

### DIFF
--- a/app/schemas.py
+++ b/app/schemas.py
@@ -22,6 +22,7 @@ class QuestionType(Enum):
     single_choice = "single-choice"
     multi_choice = "multi-choice"
     subjective = "subjective"
+    numerical = "numerical"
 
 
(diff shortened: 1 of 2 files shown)


pritamps (approved):
Looks good to me!

app/schemas.py
```diff
@@ -22,6 +22,8 @@ class QuestionType(Enum):
     single_choice = "single-choice"
     multi_choice = "multi-choice"
     subjective = "subjective"
+    integer_numerical = "integer-numerical"
```
dalmia, line 25: It would be better to use numerical-integer and numerical-float as that indicates that they are a second of categorisation.

pritamps (approved):
Looks good to me!

deepansh96 (approved):

PR #45, opened 2022-05-17
Endpoint to onboard an organization

### LINKED ISSUE #44: Create organization endpoint
- API call to create an organization.
- Auto create an API key for each org.

sudheshna-donthineni:
Fixes #44 

## Summary

- This code adds an endpoint to create a new organization. 
- When this API is called, an API key is auto created and organization is added. 
- An organization name must be sent with the API call.

Screenshot 2022-05-17 at 12 57 38 PM

P.S. Merges into the #42 PR, not main.

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -8,10 +8,23 @@ answerType = Union[List[int], str, None]
 
 class Organization(BaseModel):
     id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
-    org_name: str
+    name: str
 
     class Config:
+        allow_population_by_field_name = True
+        arbitrary_types_allowed = True
         json_encoders = {ObjectId: str}
+        schema_extra = {"example": {"name": "Avanti Fellows"}}
+
+
+class OrganizationResponse(Organization):
+    name: str
+
+    class Config:
+        allow_population_by_field_name = True
+        arbitrary_types_allowed = True
+        json_encoders = {ObjectId: str}
+        schema_extra = {"example": {"name": "Avanti Fellows"}}
 
 
 class Image(BaseModel):
--- a/app/models.py
+++ b/app/models.py
@@ -7,7 +7,11 @@ answerType = Union[List[int], str, None]
 
 
 class Organization(BaseModel):
+    id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
     org_name: str
+
+    class Config:
+        json_encoders = {ObjectId: str}
 
 
 class Image(BaseModel):
--- a/app/models.py
+++ b/app/models.py
@@ -23,7 +23,6 @@ class OrganizationResponse(Organization):
     class Config:
         allow_population_by_field_name = True
         arbitrary_types_allowed = True
-        json_encoders = {ObjectId: str}
         schema_extra = {"example": {"name": "Avanti Fellows"}}
 
 
--- a/app/routers/organizations.py
+++ b/app/routers/organizations.py
@@ -1,26 +1,43 @@
 from fastapi import APIRouter, status, HTTPException
 from database import client
-from models import Organization
+from models import Organization, OrganizationResponse
 from fastapi.responses import JSONResponse
+import secrets
 import string
-import random
+from fastapi.encoders import jsonable_encoder
 
-router = APIRouter()
+router = APIRouter(prefix="/organizations", tags=["Organizations"])
 
 
-@router.post("/create-org/{org_name}")
-async def create_organization(org_name: str):
-    if org_name is not None:
+@router.post("/", response_model=OrganizationResponse, response_model_exclude={"key"})
+async def create_organization(org: Organization):
+    org = jsonable_encoder(org)
+    if org["name"] is not None:
         # create an API key
-        char_set = string.ascii_letters + string.punctuation
-        urand = random.SystemRandom()
-        org_key = "".join([urand.choice(char_set) for _ in range(20)])
+        key = "".join(
+            [secrets.choice(string.ascii_letters + string.digits) for _ in range(20)]
+        )
+        number_of_loops = 3
+        while number_of_loops > 0:
+            # check if API key exists
+            if (client.quiz.organization.find_one({"key": key})) is None:
+                org["key"] = key
+                new_org = client.quiz.organization.insert_one(org)
+                created_org = client.quiz.organization.find_one(
+                    {"_id": new_org.inserted_id}
+                )
+                return JSONResponse(
+                    status_code=status.HTTP_201_CREATED, content=created_org
+                )
 
-        # check if API key exists
-        if (client.quiz.organization.find_one({"org_key": org_key})) is None:
-            org_data = {"org_name": org_name, "org_key": org_key}
-            client.quiz.organization.insert_one(org_data)
-            return JSONResponse(status_code=status.HTTP_201_CREATED, content=org_name)
+            else:
+                key = "".join(
+                    [
+                        secrets.choice(string.ascii_letters + string.digits)
+                        for _ in range(20)
+                    ]
+                )
+                number_of_loops -= 1
 
         raise HTTPException(
             status_code=500,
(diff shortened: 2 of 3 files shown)


app/routers/organizations.py
```diff
@@ -1,15 +1,43 @@
...
 router = APIRouter()
 
 
+@router.post("/create-org/{org_name}")
+async def create_organization(org_name: str):
+    if org_name is not None:
+        # create an API key
+        char_set = string.ascii_letters + string.punctuation
+        urand = random.SystemRandom()
+        org_key = "".join([urand.choice(char_set) for _ in range(20)])
```
pritamps, line 17: What is this line doing? Is this what is used in Plio to generate keys, @deepansh96 ?

app/routers/organizations.py
```diff
@@ -1,15 +1,43 @@
...
+
 @router.get("/check-auth-token/{api_key}", response_model=Organization)
 async def check_auth_token(api_key: str):
 
     if (
-        org := client.quiz.organization.find_one({"org_key": api_key}, {"org_name": 1})
+        org := client.quiz.organization.find_one(
+            {"org_key": api_key},
+            {
+                "org_name": 1,
```
pritamps, line 38: What is `1`? Why does it seem hardcoded?

app/routers/organizations.py
```diff
@@ -1,15 +1,43 @@
...
+@router.post("/create-org/{org_name}")
+async def create_organization(org_name: str):
+    if org_name is not None:
+        # create an API key
+        char_set = string.ascii_letters + string.punctuation
+        urand = random.SystemRandom()
+        org_key = "".join([urand.choice(char_set) for _ in range(20)])
+
+        # check if API key exists
+        if (client.quiz.organization.find_one({"org_key": org_key})) is None:
```
pritamps, line 20: Why no check for `org_name`? 

If you don't want to check for the name: since the key is generated here and not passed by the user, it doesn't make sense to return an error if the key exists. Why not just try with a new random key?

pritamps (changes requested):

app/routers/organizations.py
sudheshna-donthineni: That's the syntax. Instead of returning all the data from the query, it returns only the requested fields.
https://www.mongodb.com/docs/manual/tutorial/project-fields-from-query-results/

app/routers/organizations.py
sudheshna-donthineni: There is a check for `org_name`. 

I thought of the same thing, but I was worried about what if we reach a case where it enters an infinite loop. Won't be happening in the near future, but still.

app/routers/organizations.py
sudheshna-donthineni: In Plio, they used the secrets library, but essentially the same thing.

app/routers/organizations.py
pritamps: Add the loop with a max of 3 iterations. Break as soon as a unique id is found. No reason here to return an error when it wasn’t the user's fault

Also, use the same secrets library that Plio uses.

app/routers/organizations.py
pritamps: Ah okay cool 👌

app/routers/organizations.py
pritamps: Use the same library and a small loop as mentioned below

app/routers/organizations.py
dalmia: You won't need to do any of this in this way if you defined an appropriate OrganizationResponse model and passed that in the `response_model` argument in the decorator. This is how we have done this for all the other routes. So, let's stay consistent with that, okay?

dalmia (changes requested):

app/routers/organizations.py
pritamps: Yes yes, what @dalmia says sounds good

app/models.py
```diff
@@ -7,8 +7,12 @@
 
 
 class Organization(BaseModel):
+    id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
     org_name: str
 
+    class Config:
```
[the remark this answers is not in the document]
deepansh96: @sudheshna-donthineni take a look at this
https://pydantic-docs.helpmanual.io/usage/model_config/#options
this explains the purpose of all these config options

app/routers/organizations.py
deepansh96, line 20: You'll also have to define `org_key` in the model as well right?
Currently the model only has `org_name`

deepansh96 (changes requested):

app/routers/organizations.py
```diff
@@ -1,15 +1,43 @@
...
 from database import client
 from models import Organization
+from fastapi.responses import JSONResponse
+import string
+import random
 
 router = APIRouter()
 
 
+@router.post("/create-org/{org_name}")
```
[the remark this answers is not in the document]
sudheshna-donthineni: But we'll have to send the organization name as a param, right? How will the creator API know what the org name is?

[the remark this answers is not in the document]
sudheshna-donthineni: Actually, no, that'll be in the request model.

app/routers/organizations.py
sudheshna-donthineni: Response shouldn't have key, no?

app/routers/organizations.py
```diff
@@ -1,15 +1,57 @@
...
 
 
-@router.get("/check-auth-token/{api_key}", response_model=Organization)
+@router.post("/", response_model=OrganizationResponse)
+async def create_organization(org: Organization):
+    org = jsonable_encoder(org)
+    if org["name"] is not None:
+
+        # create an API key
+        key = "".join(
```
pritamps, line 17: I think if you move this first key generation inside the loop before the `if`, you won't need the `else` at all

pritamps (changes requested):
Just a small code thing, then good to go

app/routers/organizations.py
```diff
@@ -1,15 +1,57 @@
...
+            [secrets.choice(string.ascii_letters + string.digits) for _ in range(20)]
+        )
+        number_of_loops = 3
+
+        while number_of_loops > 0:
```
deepansh96, line 22: Also I don't really think that this extra logic is needed.
it would be very rare for a 20 character alphanumeric string to repeat itself.
and on top of that, we're just creating only one organization as of now :p 

A few things to help clean up things here
- create a separate function/method for creating random strings. Because I can see below that the API key generation code is being repeated. Why repeat the code if you can make it into a function? This function can either be here, or in some other module.
- assuming the function you created is called `generate_random_string()`, you can refactor the code as below
```py
@router.post("/", response_model=OrganizationResponse)
async def create_organization(org: Organization):
    org = jsonable_encoder(org)
   api_key = generate_random_string()
   while client.quiz.organization.find_one({"key": api_key}) is not None:
      api_key = generate_random_string()

   org["key"] = api_key
   new_org = client.quiz.organization.insert_one(org)
   created_org = client.quiz.organization.find_one(
      {"_id": new_org.inserted_id}
   )
   return created_org
```

I don't think anything else is needed.

app/routers/organizations.py
```diff
@@ -1,15 +1,57 @@
...
+                        secrets.choice(string.ascii_letters + string.digits)
+                        for _ in range(20)
+                    ]
+                )
+                number_of_loops -= 1
+
+        raise HTTPException(
+            status_code=500,
+            detail="Duplicate API key. Please try again.",
+        )
```
deepansh96, line 45: Based on my earlier comment, you can remove this
- The response message is saying "duplicate API key..." but when the backend was asked to create an organization, no API key was passed or was required.
- the post request WILL eventually create an org with API key because of the refactored code i mentioned above. so there will never be a case where we will have duplicate API key.

deepansh96 (changes requested):

deepansh96 (approved):

PR #50, opened 2022-06-02
Making metadata fields optional

### LINKED ISSUE #49: Make metadata fields optional except for quiz type

pritamps:
Fixes #49 

## Summary

All metadata fields should be optional since these will be org-specific or domain-specific

Tested by running locally and creating quiz

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -28,18 +28,18 @@ class QuizTimeLimit(BaseModel):
 
 
 class QuestionMetadata(BaseModel):
-    grade: str
-    subject: str
-    chapter: str
-    topic: str
-    competency: List[str]
-    difficulty: str
+    grade: Optional[str]
+    subject: Optional[str]
+    chapter: Optional[str]
+    topic: Optional[str]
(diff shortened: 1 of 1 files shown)


app/models.py
```diff
@@ -28,18 +28,18 @@ class QuizTimeLimit(BaseModel):
...
-    chapter: str
-    topic: str
-    competency: List[str]
-    difficulty: str
+    grade: Optional[str]
+    subject: Optional[str]
+    chapter: Optional[str]
+    topic: Optional[str]
+    competency: Optional[List[str]]
+    difficulty: Optional[str]
```
deepansh96, line 36: Don't we have to put in `= None` as well like done in classes above?

pritamps: Don't know. It worked on local without it.

I assumed `None` is just the default value if no value is set. Here if no value is set and the field itself is absent, that's okay. Unless I understand wrong!

deepansh96 (approved):

PR #56, opened 2022-06-16
Returns quiz_id in response of create_quiz POST request

### LINKED ISSUE #55: `create_quiz` POST request returns entire quiz
`create_quiz` POST request returns entire quiz. Instead return only the `quiz_id`. This reduces load on network.

suryabulusu:
Fixes #55 

## Summary
- Motivation: Returns only the `quiz_id` instead of the entire created quiz. This reduces load on network. 
- Created quiz can be retrieved by sending a GET request at `/quiz/{quiz_id}` endpoint
- Changes are also made in corresponding `app/tests` files

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -288,6 +288,19 @@ class QuizResponse(Quiz):
         }
 
 
+class QuizCreationResponse(BaseModel):
+    """Model for the response of a request that creates a quiz"""
+
+    quiz_id: str
+
+    class Config:
+        schema_extra = {
+            "example": {
+                "quiz_id": "1234",
+            }
+        }
+
+
 class SessionAnswer(BaseModel):
     """Model for the body of the request that creates a session answer"""
 
@@ -341,7 +354,7 @@ class Session(BaseModel):
 class UpdateSession(BaseModel):
     """Model for the body of the request that updates a session"""
 
-    has_quiz_ended: bool = False
+    has_quiz_ended: bool
 
     class Config:
         schema_extra = {
--- a/app/models.py
+++ b/app/models.py
@@ -224,7 +224,7 @@ class Quiz(BaseModel):
         }
 
 
-class QuizResponse(Quiz):
+class GetQuizResponse(Quiz):
     """Model for the response of any request that returns a quiz"""
 
     class Config:
@@ -288,7 +288,7 @@ class QuizResponse(Quiz):
         }
 
 
-class QuizCreationResponse(BaseModel):
+class CreateQuizResponse(BaseModel):
     """Model for the response of a request that creates a quiz"""
 
     quiz_id: str
--- a/app/models.py
+++ b/app/models.py
@@ -291,12 +291,12 @@ class GetQuizResponse(Quiz):
 class CreateQuizResponse(BaseModel):
     """Model for the response of a request that creates a quiz"""
 
-    quiz_id: str
+    id: str
 
     class Config:
         schema_extra = {
             "example": {
-                "quiz_id": "1234",
+                "id": "1234",
             }
         }
 
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -21,7 +21,7 @@ async def create_quiz(quiz: Quiz):
         client.quiz.questions.insert_many(questions)
 
     return JSONResponse(
-        status_code=status.HTTP_201_CREATED, content={"quiz_id": new_quiz.inserted_id}
+        status_code=status.HTTP_201_CREATED, content={"id": new_quiz.inserted_id}
     )
 
 
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -2,12 +2,12 @@ from fastapi import APIRouter, status, HTTPException
 from fastapi.responses import JSONResponse
 from fastapi.encoders import jsonable_encoder
 from database import client
-from models import Quiz, QuizResponse, QuizCreationResponse
+from models import Quiz, GetQuizResponse, CreateQuizResponse
 
 router = APIRouter(prefix="/quiz", tags=["Quiz"])
 
 
-@router.post("/", response_model=QuizCreationResponse)
+@router.post("/", response_model=CreateQuizResponse)
 async def create_quiz(quiz: Quiz):
     quiz = jsonable_encoder(quiz)
     new_quiz = client.quiz.quizzes.insert_one(quiz)
@@ -20,12 +20,12 @@ async def create_quiz(quiz: Quiz):
 
         client.quiz.questions.insert_many(questions)
 
-    response_content = {"quiz_id": new_quiz.inserted_id}
+    return JSONResponse(
+        status_code=status.HTTP_201_CREATED, content={"quiz_id": new_quiz.inserted_id}
+    )
 
-    return JSONResponse(status_code=status.HTTP_201_CREATED, content=response_content)
 
-
-@router.get("/{quiz_id}", response_model=QuizResponse)
+@router.get("/{quiz_id}", response_model=GetQuizResponse)
 async def get_quiz(quiz_id: str):
     if (quiz := client.quiz.quizzes.find_one({"_id": quiz_id})) is not None:
         return quiz
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -2,14 +2,13 @@ from fastapi import APIRouter, status, HTTPException
 from fastapi.responses import JSONResponse
 from fastapi.encoders import jsonable_encoder
 from database import client
-from models import Quiz, QuizResponse
+from models import Quiz, QuizResponse, QuizCreationResponse
 
 router = APIRouter(prefix="/quiz", tags=["Quiz"])
 
 
-@router.post("/")
+@router.post("/", response_model=QuizCreationResponse)
 async def create_quiz(quiz: Quiz):
-    """Returns the ID of created quiz in a dictionary with key as 'quiz_id'"""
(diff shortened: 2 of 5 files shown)


app/tests/test_sessions.py
```diff
@@ -46,15 +46,16 @@ def test_create_session_with_valid_quiz_id_and_first_session(self):
...
+        quiz = self.client.get(quizzes.router.prefix + f"/{quiz_id}").json()
         response = self.client.post(
             sessions.router.prefix + "/", json={"quiz_id": quiz["_id"], "user_id": 1}
         )
         assert response.status_code == 201
         session = json.loads(response.content)
         assert session["is_first"] is True
-        assert len(session["session_answers"]) == len(
-            quiz_data["question_sets"][0]["questions"]
+        assert len(session["session_answers"]) == sum(
```
pritamps, line 57: What is this change for?

pritamps (commented):
Looks good to me, but I will let @deepansh96 approve. I'm only starting to get familiar with quiz ka code

app/routers/quizzes.py
```diff
@@ -7,8 +7,9 @@
 router = APIRouter(prefix="/quiz", tags=["Quiz"])
 
 
-@router.post("/", response_model=QuizResponse)
+@router.post("/")
 async def create_quiz(quiz: Quiz):
+    """Returns the ID of created quiz in a dictionary with key as 'quiz_id'"""
     quiz = jsonable_encoder(quiz)
     new_quiz = client.quiz.quizzes.insert_one(quiz)
     created_quiz = client.quiz.quizzes.find_one({"_id": new_quiz.inserted_id})
```
sudheshna-donthineni, line 15: Why do we need this step anymore?

app/routers/quizzes.py
```diff
@@ -7,8 +7,9 @@
 router = APIRouter(prefix="/quiz", tags=["Quiz"])
 
 
-@router.post("/", response_model=QuizResponse)
+@router.post("/")
```
sudheshna-donthineni, line 10: Also, @deepansh96 can comment further, but instead of removing the response model, why don't we change the schema of the model.

deepansh96: Oh definitely
We should never remove the response model for a request

@suryabulusu were you able to read the pydantic docs?
This response model thing is responsible for smartly typcasting the response, checking to see if the returned response follows a particular pattern or not.
This is a very helpful layer that reduces a lot of bugs.

Removing the response model just to make the response fit into something feels like a bandage applied over the issue.

You can just change the response model no?
then the outputted response will be formatted according to that ResponseModel class

app/tests/test_sessions.py
suryabulusu: previous code assumes that there is only one question set, once we include optional we'll have more than one question set, so I edited the line to iterate over all question sets (qset)

app/routers/quizzes.py
```diff
@@ -7,8 +7,9 @@
...
 async def create_quiz(quiz: Quiz):
+    """Returns the ID of created quiz in a dictionary with key as 'quiz_id'"""
```
deepansh96, line 12: Some feedback on docstrings
- Here, the _meat_ of the method is not mentioned. The purpose of this method is not just to _return the id of the created quiz_. That is a side effect or post effect. The main task this method is doing is actually _creating the quiz_. 
- No need to mention things like `dictionary` and `with key as `quiz_id`. If you use a ResponseModel, anyone will be able to see what are the keys and what are their types that will be returned from this method call. `dictionary` is specific to python that's why you should avoid it. Plus someone reading the code, developer or not developer, should be able to understand what a method does just from the docstring

app/routers/quizzes.py
suryabulusu: response model QuizResponse is already used in GET request `get_quiz`; so I kept it as is. I could've created a QuizIdResponse model but it Pritam said just returning `id` is fine

app/routers/quizzes.py
deepansh96: Well you're correct in a way. This step is not needed anymore because we dont' need to return the whole quiz from the function's response but there are some gymnastics being applied to this returned quiz in the steps below. Basically just putting `question_set_id` inside the questions is needed and for that, created quiz is needed

app/routers/quizzes.py
```diff
@@ -20,7 +21,9 @@ async def create_quiz(quiz: Quiz):
 
         client.quiz.questions.insert_many(questions)
 
-    return JSONResponse(status_code=status.HTTP_201_CREATED, content=created_quiz)
+    return JSONResponse(
+        status_code=status.HTTP_201_CREATED, content={"quiz_id": new_quiz.inserted_id}
+    )
```
deepansh96, line 26: @suryabulusu I still feel like you haven't totally understood the purpose and usage of pydantic models
please read this well https://pydantic-docs.helpmanual.io/ 
also the models section here https://pydantic-docs.helpmanual.io/usage/models/

Or I might be wrong and you know the usage of pydantic models, but your solution is not a scalable one.
What if we need another field in the response? You'll have to add them here itself manually.
And if you remove the pydantic models, you will not get the helpful features like smart type conversion and all.
For example in this case, if you had used a response model, pydantic would automatically check if the inserted_id was a string or not, None or not and throw the exception on its own.
But if you're not using pedantic models, you'll have to write those error handlings yourself.

app/routers/quizzes.py
deepansh96: Edit - read your message on discord
You said `QuizResponse` model is already being used in `get_quiz` no?
So created another class for use in `create_quiz`?
Something like `CreatedQuizResponse` or `QuizCreationResponse`? Nothing's stopping you from creating as many model classes as you want!

app/routers/quizzes.py
deepansh96: Also, if some other dev wanted to know ki what is this function returning, they would have to comb through the code and look at the last line where you're returning JSONResponse.
Whereas in the rest of the code, all the outputs and inputs are defined as pydantic models, so everything is in one place and super easy to digest

deepansh96 (changes requested):

app/models.py
```diff
@@ -288,6 +288,19 @@ class Config:
         }
 
 
+class QuizCreationResponse(BaseModel):
+    """Model for the response of a request that creates a quiz"""
+
+    quiz_id: str
```
deepansh96, line 294: just `id` should be enough no?
It will be the id of a quiz that much is understood

app/models.py
```diff
@@ -288,6 +288,19 @@ class Config:
         }
 
 
+class QuizCreationResponse(BaseModel):
```
deepansh96, line 291: Also can you update this to `CreateQuizResponse`?
That way, you can also update the model being used in get quiz as `GetQuizResponse`.
So it will be easier to extend if we create an update method in the future.
create - CreateQuizResponse
get - GetQuizResponse
... in the future ...
update - UpdateQuizResponse etc

app/routers/quizzes.py
```diff
@@ -20,7 +20,9 @@ async def create_quiz(quiz: Quiz):
 
         client.quiz.questions.insert_many(questions)
 
-    return JSONResponse(status_code=status.HTTP_201_CREATED, content=created_quiz)
+    response_content = {"quiz_id": new_quiz.inserted_id}
+
+    return JSONResponse(status_code=status.HTTP_201_CREATED, content=response_content)
```
deepansh96, line 25:
```suggestion
    return JSONResponse(status_code=status.HTTP_201_CREATED, content={"quiz_id": new_quiz.inserted_id})
```

app/tests/base.py
```diff
@@ -22,7 +22,8 @@ def setUp(self):
         # create a question is through the quiz endpoint which is why we are using the quiz endpoint
         # to create questions and a quiz
         response = self.client.post(quizzes.router.prefix + "/", json=self.quiz_data)
-        self.quiz = json.loads(response.content)
+        self.quiz_id = json.loads(response.content)["quiz_id"]
+        self.quiz = self.client.get(quizzes.router.prefix + f"/{self.quiz_id}").json()
```
deepansh96, line 26: would need to change `quiz_id` based on my comment above

app/tests/test_quizzes.py
```diff
@@ -9,10 +9,13 @@ def setUp(self):
         self.id = self.quiz["_id"]
         self.length = len(self.quiz_data["question_sets"][0]["questions"])
 
+    def test_setup_quizId_and_quiz(self):
```
deepansh96, line 12: the name of this test method seems a little off.
Where are you setting up quiz id and quiz?

This is just verifying that the id returned from create quiz call is the same as the id of the quiz returned from get quiz call, right?
Can you try renaming it based on that?

deepansh96 (changes requested):

app/models.py
suryabulusu: let's leave it as `quiz_id` only, because creator/user will be interacting with it and it may be interpreted as `user_id`. I think in session model too we use `quiz_id`

suryabulusu: also in `router/quizzes.py`,  `@router.get("/{quiz_id})` we mention as `quiz_id`

app/models.py
suryabulusu: yes, updated

app/tests/test_sessions.py
pritamps: Ah shouldn't be a part of this PR then -- but is one line so I guess it's okay. Anyway, that's why I got confused.

PR #60, opened 2022-07-25
Model and test updates for countdown timer 

suryabulusu:
- Adds `created_at` and `updated_at` fields to session and session answer for time of submission calculations. Solves Issue #63 
- Idea is to have a global countdown timer, that begins when user first clicks `Start Quiz` on front-end. 
-- We add this event (`start-quiz`) to a session `events` list
-- Each event has a `created_at` time that records when the event occured
-- time remaining for the session in which user clicks "Start Quiz" = `quiz["time_limit"].max` (for now)
-- for subsequent sessions of the same user+quiz pair, we add the event `resume-quiz` to list when `Resume` button is clicked on frontend. 
-- Now, `time_remaining = time_remaining_in_previous_session - time_elapsed_between_current_event_and_prev_event` (explanation provided in notion doc [here](https://www.notion.so/avantifellows/Feature-Capturing-submit-time-timer-0b47c4b2fbd2405fb00cc590e27a286d))
- Tests to ensure that new session + start quiz, next session + resume quiz, etc. have different `time_remaining` values. Solves #64
- Ensure that unnecessary sessions are not being created. This involving checking two previous sessions to see if any event has occurred in previous session.  If yes, then we create a new session. If not, we can retain the previous session as new session (as it had no event)

Coresponding `quiz-frontend` PR -- https://github.com/avantifellows/quiz-frontend/pull/76


app/models.py
```diff
@@ -321,6 +323,7 @@ class UpdateSessionAnswer(BaseModel):
 
     answer: Optional[answerType]
     visited: Optional[Union[bool, None]]
+    last_updated_at: datetime = Field(default_factory=datetime.utcnow)
```
pritamps, line 326: Doesn't need to be `last_updated_at` ..you can make it just `updated_at`, it's okay :)

app/models.py
```diff
@@ -354,22 +358,37 @@ class Config:
...
+
+
 class SessionResponse(Session):
     """Model for the response of any request that returns a session"""
 
     is_first: bool
+    has_quiz_started: Optional[bool] = False
     has_quiz_ended: Optional[bool] = False
     session_answers: List[SessionAnswer]
+    quiz_start_resume_time: Optional[datetime]
```
pritamps, line 389: why `start` and `resume` in same variable name? Little weird

app/routers/sessions.py
```diff
@@ -27,13 +34,18 @@ async def create_session(session: Session):
...
 
-        # since we know that there is going to be only one question set for now
         if "question_sets" in quiz and quiz["question_sets"]:
-            question_set_id = quiz["question_sets"][0]["_id"]
-            questions = client.quiz.questions.find(
-                {"question_set_id": question_set_id}, sort=[("_id", pymongo.ASCENDING)]
-            )
+            question_set_ids = [qset["_id"] for qset in quiz["question_sets"]]
+            questions = []
+            for qset_id in question_set_ids:
```
pritamps, line 43: This doesn't belong here, no? Why is this in this PR?

app/models.py
```diff
@@ -354,22 +358,37 @@ class Config:
 class UpdateSession(BaseModel):
     """Model for the body of the request that updates a session"""
 
-    has_quiz_ended: bool
+    has_quiz_ended_first_time: bool
```
pritamps, line 361: Ended is only for last time. Maybe rename to `has_quiz_stopped_first_time`?

app/routers/sessions.py
```diff
@@ -78,30 +96,54 @@ async def create_session(session: Session):
...
+            current_time - datetime.fromisoformat(session["quiz_start_resume_time"])
+        ).seconds
+
+    if session_updates["has_quiz_ended_first_time"]:
+        # end test clicked first time
+        session["has_quiz_ended"] = True
+        session["quiz_end_time"] = current_time
+    else:
+        # quiz started or resumed
+        session["quiz_start_resume_time"] = current_time
```
pritamps, line 132: Should be separate values for `start` and `resume` (`resume` shoulnd be an array) so that we can track when the user started, stopped over multiple sessions.

Unless I'm missing something and that's already being tracked somehow

pritamps (changes requested):
I don't have much context with this code -- just some general questions

app/models.py
suryabulusu: yes

app/models.py
suryabulusu: because we can either start session or resume session when quiz opens. i'm capturing the moment when the start/resume button is clicked

app/routers/sessions.py
suryabulusu: so the `session["created"]` time will handle it. we could query for list of sessions of particular user/quiz pair - and look at array of times. however, the time captured in `created` pertains to when the quiz webpage has opened.

`start` and `resume` refer to when the user has "clicked" the start/resume button. ideally, they must be stored in an array, yes. but for current logic i didn't require. 

moreover, session start/resume times (as in, when the webpage has opened) -- can help for now.

app/routers/sessions.py
```diff
@@ -78,30 +96,54 @@ async def create_session(session: Session):
...
+
+    if session_updates["has_quiz_started_first_time"]:
+        # start/resume button clicked first time for a quiz
+        session["has_quiz_started"] = True
+        # time elapsed is zero
+    else:
+        # update time remaining based on current time (resume/end button clicked)
+        time_elapsed = (
+            current_time - datetime.fromisoformat(session["quiz_start_resume_time"])
+        ).seconds
```
deepansh96, line 124: this part is a little unclear to me.
My best guess after reading this and frontend's code is that the `else` condition is for the quiz being resumed after being paused right?

Just thinking out loud, I'll write my understanding process of this logic below.
- User comes to a quiz for the first time, a new session is created
- session['time_remaining'] is set
- user clicks "Start Quiz"
- a call to update_session API is made
- in the logic we check if the quiz is being started, resumed or ended
- if the quiz is being started and NOT resumed, we set the `quiz_start_time` in the session as the current time. time elapsed would be 0 as the quiz is just being started. so we calculate the time remaining and send it to the frontend and also update in the session document
- lets say after some time user leaves the quiz.
- they come back again, a new session is created
- this new session has details copied from the previous session. things like `has quiz started/resumed/ended` and the time remaining, quiz start time, quiz end time and quiz resume time(s). These details are outdated as they'll get updated when the user clicks resume etc.
- The user clicks resume, a call to update session API is called.
- this time the quiz is being resumed. so we'll calculate the time that has been passed since the last time session was resumed. if this is the first time session is being resumed, we'll calculate the time passed from the quiz start time
- now we have the time elapsed since the last activity (it could be quiz started or resumed)
- we'll calculate how much time is remaining since the last time_remaining and update it on DB and frontend

While writing this out the process became clear to me. Your logic is totally correct but I think it can be structured more neatly and in a way that's extendable in the future. Writing my points below so I can discuss with you on a call.

- We should segregate events. Think of starting, resuming, ending as events. Currently in your implementation, things are mixed up. ending is a separate event. starting/resuming is merged into one field. plus we're relying on the frontend to send data telling the backend that if the user is starting the quiz for the first time or not or ending the quiz for the first time or not. these are very specific events.
- Plus I think every session should contain an event string. This could be (for now) "quiz_start", "quiz_resume" and "quiz_end". Along with each event, there is a timestamp attached. My goal here is to reuse the `created_at` and `updated_at` timestamps instead of storing times in separate variables like `quiz_start_time`, `quiz_resume_time` and `quiz_end_time`.
- Going by the current implementation, In the future if we want to track some other event along with their times, we'll have to add specialized fields again. For example, if the event we want to track is `xyz`, we'll have to add `quiz_xyz_time` and then another flag let's say `has_quiz_xyz`. A better approach would be to have a general event. Which could be anything. 
- Every session will have one session-level event attached to it. And the time for the session-level event will be tracked using created_at/updated_at.
- That way, we won't have to keep track of _other_, events in one session itself. Like currently, for every session, we're tracking if the quiz was started, resumed or ended along with their times.
- With the new implementation, things will go along with the arrow of time. one session will have quiz started event with the time being tracked in created_at/updated_at. Next session will have quiz resumed event along with the time. then a few more quiz resumed events in the next few sessions. then at the end, we'll have quiz ended event. same way we can have quiz reattempted, revisited etc in the future if required.

app/models.py
deepansh96: ^this will change with the improved implementation. But the basic funda still remains valid. The events and their names should be independent!

app/models.py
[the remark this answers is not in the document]
deepansh96: ^ after going in detail through the logic, I understand what this means. The `first_time` thing essentially means whether the quiz was started or resumed.
I still didn't understand what is the meaning of `first_time` with `has_quiz_ended`. Can someone end quiz multiple times?

deepansh96 (changes requested):

app/models.py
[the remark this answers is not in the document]
suryabulusu: session answer can get updated at multiple times. `last` added to indicate we only store the latest update time. not all the update times

app/routers/sessions.py
suryabulusu: removed it from current PR

deepansh96:
- Can you update the PR description as well?
- Also create an issue and link it on the PRs (on both frontend and backend)

app/models.py
```diff
@@ -414,3 +408,12 @@ class Config:
                 ],
             }
         }
+
+
+class UpdateSessionResponse(BaseModel):
+    """Model for the response of request that updates a session"""
+
+    time_remaining: Optional[int]
```
deepansh96, line 416: add one comment here saying that this time is in seconds?

app/routers/sessions.py
```diff
@@ -59,9 +66,15 @@ async def create_session(session: Session):
...
+        if previous_session["events"] is None:
+            # later, we can avoid creating new session too
+            session["is_first"] = True
+        else:
+            session["is_first"] = False
+
+        session["events"] = previous_session.get("events", None)
         session["has_quiz_ended"] = previous_session.get("has_quiz_ended", False)
-        session["has_quiz_started"] = previous_session.get("has_quiz_started", False)
+        session["time_remaining"] = previous_session.get("time_remaining", None)
```
deepansh96, line 77: If the user didn't click `start quiz` the very first time and reloaded, we decided on _not_ to create a new session right?

app/routers/sessions.py
```diff
@@ -99,45 +109,43 @@ async def create_session(session: Session):
...
-        session["has_quiz_ended"] = True
-        session["quiz_end_time"] = current_time
-    else:
-        # quiz started or resumed
-        session["quiz_start_resume_time"] = current_time
-
     response_content = {}
     if "time_remaining" in session and session["time_remaining"] is not None:
-        # if not here => there is no time limit set, no need to update
+        # if not there => no time limit is set, no need to respond with time_remaining
```
deepansh96, line 142:
```suggestion
        # if `time_remaining` key is not present => no time limit is set, no need to respond with time_remaining
```

app/routers/sessions.py
```diff
@@ -99,45 +109,43 @@ async def create_session(session: Session):
...
+        session["events"].append(event_obj)
+
+    # diff between times of last two events
+    time_elapsed = 0
+    if new_event != EventType.start_quiz:
+        # time_elapsed = 0 for start-quiz event
         time_elapsed = (
-            current_time - datetime.fromisoformat(session["quiz_start_resume_time"])
+            str_to_datetime(session["events"][-1]["created_at"])
+            - str_to_datetime(session["events"][-2]["created_at"])
```
deepansh96, line 137: Just also add a sanity check here that the `session["events"]` array is of length of 2 or more.
I know that logically it will always be that way but still if something weird happens, I don't want the app to break

deepansh96 (changes requested):

app/models.py
```diff
@@ -159,7 +171,7 @@ class Quiz(BaseModel):
     num_graded_questions: int
     shuffle: bool = False
     num_attempts_allowed: int = 1
-    time_limit: QuizTimeLimit = None
+    time_limit: Optional[QuizTimeLimit] = None
```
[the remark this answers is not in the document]
suryabulusu: it doesn't do anything -- both lines are same but saying `Optional` makes it more explicit that this param being `None` is fine throughout the code https://www.reddit.com/r/learnpython/comments/qjtis0/pydantic_difference_between_typingoptional_and/

app/models.py
```diff
@@ -342,6 +354,7 @@ class Session(BaseModel):
     user_id: str
     quiz_id: str
     created_at: datetime = Field(default_factory=datetime.utcnow)
+    events: Optional[List[Event]]
```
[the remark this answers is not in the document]
suryabulusu: after reading a bit more on `Optional`, i think it should be `events: List[Event] = None` .. we don't intend it to be optional param, we just iniitialize with `None` to mean ki its a default value. eventually it will get populated.

app/routers/sessions.py
suryabulusu: ya i thought it could be a separate PR.. see line 70. 

because i was not sure how to handle when `resume` is not clicked.
session1 = start quiz
session2 = reload browser but don't click resume
session3 = reload browser again and click resume

 in session3, `previous_session["events"]` is not None (it has start-quiz event), but nothing happened in session2. so we could as well return session2 only. 

i felt its worthwhile to store info of cases when users just reload browsers without clicking anything, maybe useful for analysis. lets keep creating new sessions, unless we observe its a common pattern that is hurting db?

app/models.py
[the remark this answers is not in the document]
deepansh96: So I guess for more clarity we should use `Optional` along with `= None` then?
Can you make the changes everywhere where just `= None` exists without the `Optional` type?

app/models.py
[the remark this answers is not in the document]
deepansh96: Makes sense
So I guess this depends on the situation
Can you review the `models.py` file and see if any situation has an extra `Optional` which should not be there?

app/routers/sessions.py
deepansh96: I didn't understood this part, maybe you can explain on a quick call?

app/models.py
```diff
@@ -334,7 +335,7 @@ class UpdateSessionAnswer(BaseModel):
     """Model for the body of the request that updates a session answer"""
 
     answer: Optional[answerType]
-    visited: Optional[Union[bool, None]]
+    visited: Optional[bool]
```
deepansh96, line 338: is this the same as `visited: Optional[bool] = None`?

app/routers/sessions.py
```diff
@@ -25,25 +25,39 @@ def str_to_datetime(datetime_str: str) -> datetime:
...
+                "quiz_id": current_session["quiz_id"],
+                "user_id": current_session["user_id"],
+            },
+            sort=[("_id", pymongo.DESCENDING)],
+            limit=2,
+        )
     )
+    last_session, second_last_session = None, None
+    if len(previous_two_sessions) == 1:
+        # only one session exists
```
deepansh96, line 50: can you place this comment and the below comment just above the `if` and `elif` statements?

app/models.py
suryabulusu: yes.. but if `visited` is not in the `UpdateSessionAnswer` request body (sent by frontend), it gets removed -- via `remove_optional_unset_args()` in `routers/session_answers.py`

deepansh96 (approved):

PR #79, opened 2022-12-20
Adds optional feature

suryabulusu:
A new feature that gives users the opportunity to choose what questions they want to answer. 
Link to ADR: https://www.notion.so/avantifellows/ADR-5-Optional-Questions-0974a37987054d7e9d9d7a7854cd8211

Two new fields are added to QuestionSet model
1. `title` (of a question set, to display in palette)
2. `max_questions_allowed_to_attempt`
    - if this is less than `len(questions)`, the set has optional questions
    - else this has to be equal to `len(questions)` (all questions are allowed to be attempted)


app/routers/sessions.py
```diff
@@ -61,17 +61,17 @@ async def create_session(session: Session):
...
-                for question in questions:
+            question_ids = []
+            for question_set_index, question_set in enumerate(quiz["question_sets"]):
+                for question_index, question in enumerate(question_set["questions"]):
+                    question_ids.append(question["_id"])
+
+            if question_ids:
+                for question_id in question_ids:
                     session_answers.append(
                         jsonable_encoder(
```
deepansh96, line 73:
```suggestion
            question_ids = []
            for question_set_index, question_set in enumerate(quiz["question_sets"]):
                for question_index, question in enumerate(question_set["questions"]):
                    session_answers.append(
                        jsonable_encoder(
                             SessionAnswer.parse_obj({ "question_id": question["_id"] })
                        )
                    )
```

Why not do it at once?
There's no need of two separate loops.

app/tests/base.py
```diff
@@ -25,15 +25,6 @@ def setUp(self):
...
 
-        # long assessment quiz
-        self.long_assessment_quiz_data = json.load(
-            open("app/tests/dummy_data/long_assessment_quiz.json")
-        )
-        (
-            self.long_assessment_quiz_id,
-            self.long_assessment_quiz,
-        ) = self.post_and_get_quiz(self.long_assessment_quiz_data)
-
```
deepansh96, line 36: why did you remove this? was this test data updated with multi question set data?

deepansh96 (changes requested):

app/tests/base.py
suryabulusu: yes, whatever tests we were making with long assessment, we can do the same with multi question set quiz. hence, i removed it. i've retained the long assessment json file though.

app/routers/sessions.py
suryabulusu: correct, don't know why i made two loops

app/models.py
```diff
@@ -162,6 +162,9 @@ class Config:
 class QuestionSet(BaseModel):
     id: PyObjectId = Field(default_factory=PyObjectId, alias="_id")
     questions: List[Question]
+    title: Optional[str]
+    max_questions_allowed_to_attempt: int
+    # optional questions exist if max_questions_allowed_to_attempt not equal to len(questions)
```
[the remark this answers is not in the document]
suryabulusu: we will ensure in `cms-to-quiz-engine` script code that the key is always filled with a value (mostly `len(questions)`). and yes we'll need to update all question sets in DB with key,val pair. 

I was trying to set a default value (`len(questions)`) using validators or something via pydantic . But it didn't work. 

Why avoid updating DB?

Issue #82, opened 2023-03-28
The `session-answers` table is redundant as all that information is also duplicated in the individual `sessions`


deepansh96:

suryabulusu:
Handled in #84

PR #84, opened 2023-03-28
Removed dependency on redundant `session_answers` table + optimized some mongo queries

### LINKED ISSUE #83: At some places we're updating just one key of an object or one element of an array but we send the whole object to MongoDB to update which is inefficient.

deepansh96:
Fixes #83 #82 
ADR - https://www.notion.so/avantifellows/ADR-Quiz-Backend-DB-optimization-66e73d0531854098a2702fee070e2fc5

### DIFF
--- a/app/routers/session_answers.py
+++ b/app/routers/session_answers.py
@@ -8,50 +8,50 @@ from utils import remove_optional_unset_args
 router = APIRouter(prefix="/session_answers", tags=["Session Answers"])
 
 
-@router.patch("/{session_answer_id}", response_model=SessionAnswerResponse)
-async def update_session_answer(
-    session_answer_id: str, session_answer: UpdateSessionAnswer
+@router.patch("/{session_id}/{position_index}", response_model=SessionAnswerResponse)
+async def update_session_answer_in_a_session(
+    session_id: str, position_index: int, session_answer: UpdateSessionAnswer
 ):
+    """
+    Update a session answer in a session by its position index in the session answers array
+    Path Params:
+    session_id - the id of the session
+    position_index - the position index of the session answer in the session answers array. This corresponds to the position of the question in the quiz
+    """
     session_answer = remove_optional_unset_args(session_answer)
     session_answer = jsonable_encoder(session_answer)
 
-    if (client.quiz.session_answers.find_one({"_id": session_answer_id})) is None:
+    # check if the session exists
+    session = client.quiz.sessions.find_one({"_id": session_id})
+    if session is None:
         raise HTTPException(
             status_code=status.HTTP_404_NOT_FOUND,
-            detail=f"session_answer {session_answer_id} not found",
+            detail=f"Provided session with id {session_id} not found",
         )
+
+    # check if the session has session answers key
+    if "session_answers" not in session or session["session_answers"] is None:
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"No session answers found in the session with id {session_id}",
+        )
+
+    # check if the session answer index that we're trying to access is out of bounds or not
+    if position_index > len(session["session_answers"]):
+        raise HTTPException(
+            status_code=status.HTTP_400_BAD_REQUEST,
+            detail=f"Provided position index {position_index} is out of bounds of length of the session answers array",
+        )
+
+    # constructing the $set query for mongodb
+    setQuery = {}
+    for key, value in session_answer.items():
+        setQuery[f"session_answers.{position_index}.{key}"] = value
 
     # update the document in the session_answers collection
-    client.quiz.session_answers.update_one(
-        {"_id": session_answer_id}, {"$set": session_answer}
-    )
+    client.quiz.sessions.update_one({"_id": session_id}, {"$set": setQuery})
 
-    updated_session_answer = client.quiz.session_answers.find_one(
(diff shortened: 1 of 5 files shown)


app/routers/session_answers.py
```diff
@@ -8,62 +8,76 @@
 router = APIRouter(prefix="/session_answers", tags=["Session Answers"])
 
 
-@router.patch("/{session_answer_id}", response_model=SessionAnswerResponse)
-async def update_session_answer(
-    session_answer_id: str, session_answer: UpdateSessionAnswer
+@router.patch("/{session_id}/{position_index}", response_model=SessionAnswerResponse)
+async def update_session_answer_in_a_session(
+    session_id: str, position_index: int, session_answer: UpdateSessionAnswer
```
pritamps, line 13: Ah cool :)

app/routers/session_answers.py
```diff
@@ -8,62 +8,76 @@
...
+            detail=f"No session answers found in the session with id {session_id}",
+        )
 
-    # update the document in the sessions collection if this answer
-    # is present in the subset of session answers we store in the document
-    # corresponding to the session
-    session_to_update = client.quiz.sessions.find_one(
-        {"_id": updated_session_answer["session_id"]}
-    )
+    # check if the session answer index that we're trying to access is out of bounds or not
```
pritamps, line 39: Is there a possibility of this error coming up, or this check is just for safety?

app/routers/session_answers.py
```diff
@@ -8,62 +8,76 @@
...
-    if update_session:
-        client.quiz.sessions.update_one(
-            {"_id": session_to_update["_id"]},
-            {"$set": {"session_answers": session_answers}},
-        )
+    # update the document in the session_answers collection
+    client.quiz.sessions.update_one({"_id": session_id}, {"$set": setQuery})
 
-    return JSONResponse(status_code=status.HTTP_200_OK, content=updated_session_answer)
+    return JSONResponse(status_code=status.HTTP_200_OK, content=session_answer)
```
pritamps, line 54: Do we need to return the `session_answer` for frontend? Or just returning `OK` will be enough?

pritamps (commented):
Looks good to me. Just questions for my own context building!

app/routers/session_answers.py
suryabulusu: I don't think it is possible. When we create a session, we also create an array of session answers with size `len(questions)`. Position index will be bound by the same size, so we won't access out of bounds. But good safety check

app/routers/session_answers.py
suryabulusu: we don't use the response anywhere. but i'm a little bit confused wrt response model `SessionAnswerResponse`. It needs to have a `session_id` key which is not there in current response. Will get back here after reading entire code

app/routers/session_answers.py
```diff
@@ -8,62 +8,76 @@
...
+            detail="Either session_id is wrong or position_index is out of bounds",
         )
-    ) is not None:
-        return session_answer
 
-    raise HTTPException(
-        status_code=status.HTTP_404_NOT_FOUND,
-        detail=f"session_answer {session_answer_id} not found",
+    return JSONResponse(
+        status_code=status.HTTP_200_OK, content=aggregation_result[0]["session_answer"]
```
suryabulusu, line 82: does this content response contain `session_id` key? which is supposed to be there in `response_model` for this function

app/routers/sessions.py
```diff
@@ -120,17 +116,10 @@ async def create_session(session: Session):
...
-    # update with new session_id and insert to db
-    for index, _ in enumerate(session_answers):
-        session_answers[index]["session_id"] = new_session.inserted_id
-
-    client.quiz.session_answers.insert_many(session_answers)
+    client.quiz.sessions.insert_one(current_session)
 
     # return the created session
-    return JSONResponse(status_code=status.HTTP_201_CREATED, content=created_session)
+    return JSONResponse(status_code=status.HTTP_201_CREATED, content=current_session)
```
suryabulusu, line 122: `current_session` doesn't have the new `_id` corresponding to `client.quiz.sessions.insert_one(current_session)` na.. we need to send the object after insertion. I think we need to retain `created_session` variable

app/routers/session_answers.py
deepansh96: Have updated. Returning nothing back to frontend and removed the response model

app/routers/session_answers.py
deepansh96: Removed the response model from this. This is anyway just used by the backend tests and not used anywhere in the frontend

app/routers/sessions.py
```diff
@@ -212,15 +204,14 @@ async def update_session(session_id: str, session_updates: UpdateSession):
     ):
         # if `time_remaining` key is not present =>
         # no time limit is set, no need to respond with time_remaining
-        session["time_remaining"] = max(0, session["time_remaining"] - time_elapsed)
+        time_remaining = max(0, session["time_remaining"] - time_elapsed)
+        session_update_query["$set"] = {"time_remaining": time_remaining}
         response_content = {"time_remaining": session["time_remaining"]}
```
suryabulusu, line 209: should be `response_content = {"time_remaining": time_remaining]}` since we haven't updated in session yet

PR #88, opened 2023-04-13
Omr feature

### LINKED ISSUE #89: Add an OMR view for quesiton player
A new view that allows users to key-in their answers quickly on a single page. This is useful when users solve quizzes offline.

suryabulusu:
Fixes #89 

## Summary

This PR contains changes to enable OMR view of question player. 
- Adds a new value to `QuizType` in `schemas.py`
- When returning an omr quiz, questions beyond subset size within a question set are expected to contain an array of options instead of empty array. The size of the array is equal to number of options for that question. This lets frontend know the number of options to display for every question. Contents of the array are immaterial for OMR view.
- Adds tests

Related ADR: https://www.notion.so/avantifellows/ADR-OMR-Quiz-Type-61dde7c99e4540d0b7ef89350753abbc


app/routers/quizzes.py
```diff
@@ -92,6 +93,35 @@ async def get_quiz(quiz_id: str):
     quiz_collection = client.quiz.quizzes
     if (quiz := quiz_collection.find_one({"_id": quiz_id})) is not None:
         update_quiz_for_backwards_compatibility(quiz_collection, quiz_id, quiz)
+
+        if quiz["metadata"]["quiz_type"] == QuizType.omr.value:
+            print("true", QuizType.omr)
```
deepansh96, line 98: remove this?

app/routers/quizzes.py
```diff
@@ -92,6 +93,35 @@ async def get_quiz(quiz_id: str):
     quiz_collection = client.quiz.quizzes
     if (quiz := quiz_collection.find_one({"_id": quiz_id})) is not None:
         update_quiz_for_backwards_compatibility(quiz_collection, quiz_id, quiz)
+
+        if quiz["metadata"]["quiz_type"] == QuizType.omr.value:
```
deepansh96, line 97: also put null checks for these keys no?
Like if `metadata` is not in quiz or `quiz_type` is not in `quiz["metadata"]`? Just to be sure?

app/routers/quizzes.py
```diff
@@ -92,6 +93,35 @@ async def get_quiz(quiz_id: str):
...
+            print("true", QuizType.omr)
+            for question_set_index, question_set in enumerate(quiz["question_sets"]):
+                options_count_aggregate = client.quiz.questions.aggregate(
+                    [
+                        {"$match": {"question_set_id": question_set["_id"]}},
+                        {"$sort": {"_id": 1}},
+                        {"$skip": settings.subset_size},
+                        {"$project": {"number_of_options": {"$size": "$options"}}},
+                    ]
+                )
```
deepansh96, line 107: You're running this aggregation query 4 times (for 4 question sets). Everytime mongo has to parse through ALL the questions in the collection of the questions 4 times. 
You can club this into one aggregation query.
```
question_set_ids = [ q_set["_id"] for q_set in quiz["question_sets"] ]
options_count_aggregate = client.quiz.questions.aggregate(
    [
        {"$match": {"question_set_id": {
            "$in": question_set_ids
        }}},
        {"$sort": {"_id": 1}},
        {"$skip": settings.subset_size},
        {"$project": {"number_of_options": {"$size": "$options"}}},
    ]
)
```

Now you have option_counts for all the questions belonging to all the question_sets in one array.
You should also project `question_set_id` in this case so you can parse and store it

deepansh96: Also add a comment on top of this no, what are you trying to do here.
Took some time to understand what information will `options_count_aggregate` will store

app/routers/quizzes.py
```diff
@@ -92,6 +93,35 @@ async def get_quiz(quiz_id: str):
...
+                options_count = [
+                    doc["number_of_options"] for doc in list(options_count_aggregate)
+                ]
```
deepansh96, line 110: This is not at all necessary. This is an extra loop over all the values where you're just extracting `number_of_options` from the array of objects. You can add two stages to your aggregation query to return you the array of number directly instead of array of objects.
This is how the final query might look like
```
[
  {
      '$match': {
          'question_set_id': '64344359edcb8c78aef1ce1e'
      }
  }, {
      '$sort': {
          '_id': 1
      }
  }, {
      '$skip': 10
  }, {
      '$project': {
          '_id': 0, 
          'number_of_options': {
              '$size': '$options'
          }
      }
  }, {
      '$group': {
          '_id': None, 
          'optionsArray': {
              '$push': '$number_of_options'
          }
      }
  }, {
      '$project': {
          '_id': 0, 
          'optionsArray': 1
      }
  }
]
```

(chat gpt helped me with this :p)

deepansh96 (changes requested):

app/routers/quizzes.py
```diff
@@ -92,6 +93,75 @@ async def get_quiz(quiz_id: str):
     quiz_collection = client.quiz.quizzes
     if (quiz := quiz_collection.find_one({"_id": quiz_id})) is not None:
         update_quiz_for_backwards_compatibility(quiz_collection, quiz_id, quiz)
+
+        if "metadata" in quiz and quiz["metadata"] is not None:
+            if "quiz_type" in quiz["metadata"]:
```
pritamps, line 98: So we're assuming that if quiz is not `omr-type`, then assume it's normal quiz? Makes sense

app/routers/quizzes.py
```diff
@@ -92,6 +93,75 @@ async def get_quiz(quiz_id: str):
...
+                        options_count_per_set = options_count_across_sets[
+                            question_set_index
+                        ]["options_count_per_set"]
+                        for question_index, question in enumerate(
+                            question_set["questions"]
+                        ):
+                            if question_index < settings.subset_size:
+                                continue
+
+                            # options_count will be zero for subbjective/numerical questions
```
pritamps, line 155: Why? This is a behaviour of the current test design? Maybe the quiz engine shouldn't depend on it, no?

app/schemas.py
```diff
@@ -39,6 +39,7 @@ class QuizLanguage(Enum):
 class QuizType(Enum):
     assessment = "assessment"
     homework = "homework"
+    omr = "omr-assessment"
```
pritamps, line 42: This is a bit weird. There is a lot of behaviour difference between `assessment` and `homework`, but OMR will now be supported only for one sort of behaviour (`assessment`). 

Is that what we want?

pritamps (commented):
Just some comments

app/routers/quizzes.py
suryabulusu: yes

app/routers/quizzes.py
suryabulusu: Do you mean `options_count` being zero for subjective/numerical? Quiz Engine doesn't depend on it as such. The above mongo aggregate will set it to zero anyway, as in, Line158 `options_count_per_set[question_index]` would be zero for a subjective/numerical question.

The reason we are doing this aggregate: our subset pattern logic doesn't send number-of-options for questions beyond subset size. Omr view requires info of number-of-options for all questions to render its components. 

Had we not done this aggregate, questions beyond subset size would look blank in `OmrModal`, not displaying any options (bubbles).

app/schemas.py
suryabulusu: Yes. OMR flow makes sense only for assessments.
For homeworks, once a question is answered, we cannot change answer again. I wonder if that would look good for a "omr-homework" quiz type. What if user accidentally touches a bubble, etc.

Also homeworks are for at-home learning. We can assume students will have enough time to go through questions.

We also send fewer questions in our homeworks currently (like 5-10). It should be easy to navigate across them quickly in question player view.

deepansh96 (approved):

PR #91, opened 2023-05-07
Remove extra dummy events

suryabulusu:
Production test session ID: `633646b0639829dad5a896b4`

Fixes #{issue id}


app/routers/sessions.py
```diff
@@ -154,11 +154,34 @@ async def update_session(session_id: str, session_updates: UpdateSession):
...
+            new_event == EventType.dummy_event
+            and session["events"][-1]["event_type"] == EventType.dummy_event
+        ):
+            # if previous event is dummy, just change the updated_at time of previous event
+            last_event_index = len(session["events"]) - 1
+            if "$set" not in session_update_query:
+                session_update_query["$set"] = {
+                    "events."
+                    + str(last_event_index)
+                    + ".updated_at": event_obj["created_at"]
```
deepansh96, line 168: When we get the new event, we're calling it `new_event` and we do this `new_event = jsonable_encoder(session_updates)["event"]`.

Then we're creating a structured object out of it by doing
`event_obj = jsonable_encoder(Event.parse_obj({"event_type": new_event}))`

We should also name it as `new_event_obj` instead of `event_obj` just to keep things crystal clear

app/routers/sessions.py
```diff
@@ -154,11 +154,34 @@ async def update_session(session_id: str, session_updates: UpdateSession):
...
+                }
+            else:
+                session_update_query["$set"].update(
+                    {
+                        "events."
+                        + str(last_event_index)
+                        + ".updated_at": event_obj["created_at"]
+                    }
+                )
```
deepansh96, line 177:
```suggestion
            query = {
                    "events."
                    + str(last_event_index)
                    + ".updated_at": event_obj["created_at"]
            }
            if "$set" not in session_update_query:
                session_update_query["$set"] = query
            else:
                session_update_query["$set"].update(query)
```

app/scripts/remove_extra_dummy_events_in_sessions.py
```diff
@@ -0,0 +1,69 @@
...
+    if "MONGO_AUTH_CREDENTIALS" not in os.environ:
+        from dotenv import load_dotenv
+
+        load_dotenv("../.env.local")
+    client = MongoClient(os.getenv("MONGO_AUTH_CREDENTIALS"))
+    session_collection = client.quiz.sessions
+
+    # Iterate through all documents with an 'events' key
+    for session in session_collection.find(
+        {"events": {"$exists": True}, "$expr": {"$gt": [{"$size": "$events"}, 1000]}}
```
deepansh96, line 17: why are we only filtering where size of events array is more than 1000?

suryabulusu: `session_collection.find({"events": {"$exists": True}})` I'll change it to this. Large event arrays will definitely contain dummy-events, so I filtered for one of those sessions to check how script is running.

app/routers/sessions.py
suryabulusu: done

PR #96, opened 2023-06-13
set up logger and log statements

deepansh96:
## Summary

The setup logger here will log the statements to stdout. These will then appear in cloudwatch logs. A separate lambda (lambda-promtail) will be configured to listen to the cloudwatch logs of quiz backend and forward them to loki.


app/logger_config.py
```diff
@@ -0,0 +1,32 @@
...
+def setup_logger():
+    logger = logging.getLogger("quizenginelogger")
+
+    logger_format = "%(asctime)s IST loglevel=%(levelname)-6s filename=%(filename)s funcName=%(funcName)s() L%(lineno)-4d %(message)s call_trace=%(pathname)s L%(lineno)-4d"
+
+    formatter = ISTFormatter(fmt=logger_format, datefmt="%Y-%m-%d %H:%M:%S")
+
+    consoleHandler = logging.StreamHandler()
+    consoleHandler.setLevel(logging.DEBUG)
+    consoleHandler.setFormatter(formatter)
```
pritamps, line 24: `consoleHandler` doesn't seem to be returned from here. What's it used for?

pritamps (approved):
Looks good to me!

app/logger_config.py
deepansh96: no need to return. We've already added the console handler as a `handler` to the logger. We just need to return the logger, not the handler.

pritamps (approved):

app/main.py
```diff
@@ -1,13 +1,45 @@
...
+from logger_config import setup_logger
+
+logger = setup_logger()
 
 COMPRESS_MIN_THRESHOLD = 1000  # if more than 1000 bytes (~1KB), compress
 
 app = FastAPI()
 
+
+@app.middleware("http")
```
suryabulusu, line 18: why only http, i think our quiz runs on https

app/routers/questions.py
```diff
@@ -30,8 +38,12 @@ async def get_questions(question_set_id: str, skip: int = None, limit: int = Non
         pipeline.append({"$limit": limit})
 
     if (questions := list(client.quiz.questions.aggregate(pipeline))) is not None:
+        logger.info(
+            f"Found {len(questions)} questions with question_set_id: {question_set_id}"
+        )
         return questions
 
+    logger.error("No questions found")
```
suryabulusu, line 46: logger.error("No questions found with question_set_id: {question_set_id}")

app/routers/quizzes.py
```diff
@@ -39,19 +44,42 @@ def update_quiz_for_backwards_compatibility(quiz_collection, quiz_id, quiz):
...
+            status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+            detail="Failed to update quiz for backwards compatibility",
+        )
+
+    logger.info("Quiz updated for backwards compatibility")
 
 
 @router.post("/", response_model=CreateQuizResponse)
 async def create_quiz(quiz: Quiz):
+    logger.info("Starting quiz creation")
```
suryabulusu, line 66: can we have some description for quiz? 

we could use quiz source id for now. its in metadata
we should do this after `jsonable_encoder(quiz)` step

if "metadata" in quiz and "source-id" in quiz["metadata"]:
    logger.info("starting quiz creation with source id quiz["metadata"]["source_id"]

app/routers/quizzes.py
```diff
@@ -39,19 +44,42 @@ def update_quiz_for_backwards_compatibility(quiz_collection, quiz_id, quiz):
...
 
-        client.quiz.questions.insert_many(questions)
+        result = client.quiz.questions.insert_many(questions)
+        if result.acknowledged:
+            logger.info("Inserted questions for quiz")
+        else:
+            logger.error("Failed to insert questions for quiz")
+            raise HTTPException(
+                status_code=status.HTTP_500_INTERNAL_SERVER_ERROR,
+                detail="Failed to insert questions for quiz",
```
suryabulusu, line 81: add source_id here too if it exists

app/routers/quizzes.py
```diff
@@ -81,89 +109,97 @@ async def create_quiz(quiz: Quiz):
         aggregated_questions = list(subset_with_details) + list(subset_without_details)
         quiz["question_sets"][question_set_index]["questions"] = aggregated_questions
 
-    new_quiz = client.quiz.quizzes.insert_one(quiz)
+    new_quiz_result = client.quiz.quizzes.insert_one(quiz)
+    if not new_quiz_result.acknowledged:
+        logger.error("Failed to insert quiz")
```
suryabulusu, line 114: source_id

app/routers/session_answers.py
```diff
@@ -18,26 +20,34 @@ async def update_session_answer_in_a_session(
     session_id - the id of the session
     position_index - the position index of the session answer in the session answers array. This corresponds to the position of the question in the quiz
     """
+    logger.info(
```
suryabulusu, line 23: is session_answer.answer accessible here before doing jsonable_encoder

app/routers/session_answers.py
```diff
@@ -18,26 +20,34 @@ async def update_session_answer_in_a_session(
...
+        f"Updating session answer for session: {session_id} at position: {position_index}. The answer is {session_answer.answer}"
```
suryabulusu, line 24: is session_answer.answer accessible here before doing jsonable_encoder(session_answer)

suryabulusu: also, sometimes, session_answer.answer won't change. but only `session_answer.visited` is set to True. we can add this also to logger

app/routers/sessions.py
```diff
@@ -21,15 +22,20 @@ def str_to_datetime(datetime_str: str) -> datetime:
 
 
 router = APIRouter(prefix="/sessions", tags=["Sessions"])
+logger = get_logger()
 
 
 @router.post("/", response_model=SessionResponse)
 async def create_session(session: Session):
+    logger.info(
```
suryabulusu, line 30: do we have access to session.user_id here before doing `jsonable_encoder(session)`

app/routers/sessions.py
```diff
@@ -21,15 +22,20 @@ def str_to_datetime(datetime_str: str) -> datetime:
...
+        f"Creating new session for user: {session.user_id} and quiz: {session.quiz_id}"
+    )
     current_session = jsonable_encoder(session)
 
     quiz = client.quiz.quizzes.find_one({"_id": current_session["quiz_id"]})
 
     if quiz is None:
+        logger.error(f"Quiz {current_session['quiz_id']} not found")
```
suryabulusu, line 38: logger.error(f"Quiz {current_session['quiz_id']} not found while creating session") ?

app/routers/sessions.py
```diff
@@ -87,13 +96,19 @@ async def create_session(session: Session):
         )
 
         if condition_to_return_last_session is True:
+            logger.info(
```
suryabulusu, line 99: returning `last_session` with id `last_session["_id"]`

app/routers/sessions.py
```diff
@@ -131,6 +154,7 @@ async def update_session(session_id: str, session_updates: UpdateSession):
     * end button is clicked (end-quiz event)
     * dummy event logic added for JNV -- will be removed!
     """
+    logger.info(f"Updating session with id {session_id}")
```
suryabulusu, line 157: i think we can include what `session_updates` we received too. 
`new_event = jsonable_encoder(session_updates)["event"]`

suryabulusu: it will always be one of dummy / start / resume / end

app/main.py
deepansh96: https is handled by a separate layer, not this web server. this web server only sees, responds, and communicates using HTTP. Https is handled by a TLS layer which this web server doesn't have access to. The encryption and decryption of data packets is not handled by this web server.

app/routers/session_answers.py
deepansh96: For the first comment, yes we can access it

app/routers/sessions.py
deepansh96: yes

PR #98, opened 2023-07-13
Update all session answers at once

suryabulusu:
When a patch request with only `session_id` is provided to `session_answers` route, assert that json body contains all session answers, and update all session answers at once in db. 

This helps in syncing all session answers from frontend before test ends (especially for OMR format in low network areas). 

One test added in `test_session_answers.py`

Solves #99 

Related ADR: https://www.notion.so/avantifellows/ADR-Wait-for-Session-Answer-Response-ce48a939beb240a5b56f5c6035c53931

### DIFF
--- a/app/routers/session_answers.py
+++ b/app/routers/session_answers.py
@@ -89,7 +89,7 @@ async def update_all_session_answers_in_a_session(
     session_id: str, session_answers: List[UpdateSessionAnswer]
 ):
     """
-    Update all session answers in a session in the session answers array
+    Update all answers at once in the `session_answers` array
     Path Params:
     session_id - the id of the session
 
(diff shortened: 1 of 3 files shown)


app/routers/session_answers.py
```diff
@@ -83,6 +84,84 @@ async def update_session_answer_in_a_session(
     return JSONResponse(status_code=status.HTTP_200_OK, content=None)
 
 
+@router.patch("/{session_id}", response_model=None)
+async def update_all_session_answers_in_a_session(
+    session_id: str, session_answers: List[UpdateSessionAnswer]
+):
+    """
+    Update all session answers in a session in the session answers array
```
deepansh96, line 92: weird grammar.

app/routers/session_answers.py
```diff
@@ -83,6 +84,84 @@ async def update_session_answer_in_a_session(
...
+
+    # check if the session exists
+    session = client.quiz.sessions.find_one({"_id": session_id})
+    if session is None:
+        logger.error(
+            f"Received all session_answer update request, but provided session with id {session_id} not found"
+        )
+        raise HTTPException(
+            status_code=status.HTTP_404_NOT_FOUND,
+            detail=f"Received all session_answer update request, but provided session with id {session_id} not found",
```
deepansh96, line 113: you can save this in a variable and use it in the logger.error above and here as well. No need to write the log statements strings again

app/routers/session_answers.py
```diff
@@ -83,6 +84,84 @@ async def update_session_answer_in_a_session(
...
+        )
+
+    # get user_id and quiz_id for logging
+    # Note: every session must have these keys
+    user_id, quiz_id = session["user_id"], session["quiz_id"]
+    log_message += f"(user: {user_id}, quiz: {quiz_id})"
```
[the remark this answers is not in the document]
suryabulusu: yes ... its a `+=` . `log_message` variable already has session_id thing

PR #113, opened 2024-04-22
Adds metrics model

suryabulusu:
- Frontend metrics are received in backend and sent to mongoDB
- Added to sessions collection document

### DIFF
--- a/app/models.py
+++ b/app/models.py
@@ -73,23 +73,10 @@ class Event(BaseModel):
     updated_at: datetime = Field(default_factory=datetime.utcnow)
 
 
-class MetricPayload(BaseModel):
-    # metric object received from frontend
-    name: str
-    marksScored: int
-    maxQuestionsAllowedToAttempt: int
-    numAnswered: int
-    correctlyAnswered: int
-    wronglyAnswered: int
-    partiallyAnswered: int
-    attemptRate: float
-    accuracyRate: float
-
-
 class QuestionSetMetric(BaseModel):
     name: str
+    qset_id: str
     marks_scored: int
-    max_questions_allowed_to_attempt: int
     num_answered: int
     num_skipped: int
     num_correct: int
@@ -100,9 +87,7 @@ class QuestionSetMetric(BaseModel):
 
 
 class SessionMetrics(BaseModel):
-    total_sets: int
     qset_metrics: List[QuestionSetMetric]
-    total_questions: int
     total_answered: int
     total_skipped: int
     total_correct: int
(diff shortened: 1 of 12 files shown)


app/models.py
```diff
@@ -73,6 +73,44 @@ class Event(BaseModel):
...
+    wronglyAnswered: int
+    partiallyAnswered: int
+    attemptRate: float
+    accuracyRate: float
+
+
+class QuestionSetMetric(BaseModel):
+    name: str
+    marks_scored: int
+    max_questions_allowed_to_attempt: int
```
deepansh96, line 92: Why is this needed from frontend?

app/models.py
```diff
@@ -73,6 +73,44 @@ class Event(BaseModel):
...
+    num_skipped: int
+    num_correct: int
+    num_wrong: int
+    num_partially_correct: int
+    attempt_rate: float
+    accuracy_rate: float
+
+
+class SessionMetrics(BaseModel):
+    total_sets: int
```
deepansh96, line 103: what is sets here? question sets? what does this metric represent about a session?

app/models.py
```diff
@@ -423,6 +461,7 @@ class Session(BaseModel):
     created_at: datetime = Field(default_factory=datetime.utcnow)
     events: List[Event] = []
     has_quiz_ended: bool = False
+    metrics: Optional[SessionMetrics] = None
```
deepansh96, line 464: why is this needed? when session starts there will be no metrics to report right?

app/routers/sessions.py
```diff
@@ -117,6 +170,7 @@ async def create_session(session: Session):
         current_session["events"] = last_session.get("events", [])
         current_session["time_remaining"] = last_session.get("time_remaining", None)
         current_session["has_quiz_ended"] = last_session.get("has_quiz_ended", False)
+        current_session["metrics"] = last_session.get("metrics", None)
```
deepansh96, line 173: so we're carrying over the metrics across sessions?

app/routers/sessions.py
```diff
@@ -274,10 +330,17 @@ async def update_session(session_id: str, session_updates: UpdateSession):
 
     # update the document in the sessions collection
     if new_event == EventType.end_quiz:
+        session_metrics = reformat_metrics(jsonable_encoder(session_updates)["metrics"])
```
deepansh96, line 333: one thing I didn't understand is why are we doing this reformatting / extra processing. Isn't everything already calculated on the frontend and should be just sent to the DB through the backend?
Also worried about these extra processing because if its unnecessary computation then it will put load on the system when its under scale. Either if its lambda or ec2 server.

deepansh96 (changes requested):

app/models.py
suryabulusu: i will remove it

app/models.py
suryabulusu: no need to remove it

app/routers/sessions.py
suryabulusu: removed reformatting. shifted to frontend

deepansh96 (approved):

PR #121, opened 2025-01-26
Shuffle Question Order in Quiz Assessments/Homeworks

Asp-irin:
## **Description**  
This PR enhances the quiz assessment feature by introducing randomized question shuffling and ensuring accurate response storage under shuffled conditions.  

- **Logic Overview**:
  - Questions are displayed in a **randomized order**.
  - Each displayed question is mapped to its **original index**.
  - User responses are saved at the **original index** in the backend.

#### **Key Highlights**  
- [x] **Shuffling Logic:** Applied only when the quiz is not of OMR type.  
- [x] **Randomized Subsets:** Carefully maintains the balance between subset size and total question set size.  
- [x] **Index Management:** Ensures responses are stored accurately despite shuffling.  
- [x] **Session Storage:**  The random order from the first session is stored and reused across all sessions of the same quiz, ensuring consistency and preventing confusion for the user.

### DIFF
--- a/app/routers/sessions.py
+++ b/app/routers/sessions.py
@@ -25,12 +25,6 @@ def str_to_datetime(datetime_str: str) -> datetime:
 
 
 def shuffle_question_order(question_sets):
-    # Assuming `state['question_sets']` is a list of question arrays (list of lists) and `state['bucket_size']` is an integer
-    # state = {
-    #     'question_sets': question_sets,  # Replace with your actual question sets
-    #     'bucket_size': bucket_size,  # Replace with the actual bucket size
-    #     'question_order': []  # This will hold the shuffled question order
-    # }
     question_order = []
     bucket_size = Settings().subset_size
 
@@ -71,7 +65,7 @@ logger = get_logger()
 @router.post("/", response_model=SessionResponse)
 async def create_session(session: Session):
     logger.info(
-        f"Creating new session for user: {session.user_id} and quiz: {session.quiz_id} and {session.omr_mode}"
+        f"Creating new session for user: {session.user_id} and quiz: {session.quiz_id}"
     )
     current_session = jsonable_encoder(session)
 
@@ -98,8 +92,6 @@ async def create_session(session: Session):
             limit=2,
         )
     )
-    logger.info(f"Found {(previous_two_sessions)} previous sessions")
-
     last_session, second_last_session = None, None
     # only one session exists
     if len(previous_two_sessions) == 1:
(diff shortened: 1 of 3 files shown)


app/routers/sessions.py
```diff
@@ -136,6 +175,7 @@ async def create_session(session: Session):
         current_session["time_remaining"] = last_session.get("time_remaining", None)
         current_session["has_quiz_ended"] = last_session.get("has_quiz_ended", False)
         current_session["metrics"] = last_session.get("metrics", None)
+        current_session["question_order"] = last_session["question_order"]
```
suryabulusu, line 178: you sure this shouldn't have a default value here? i'm wondering if it will cause issues. with previously completed quizzes, etc. this necessitates that we add identity order to all existing sessions, right?

app/routers/sessions.py
```diff
@@ -70,6 +106,9 @@ async def create_session(session: Session):
     if last_session is None:
         logger.info("No previous session exists for this user-quiz combo")
         current_session["is_first"] = True
+        if not session.omr_mode:
+            question_order = shuffle_question_order(quiz["question_sets"])
```
suryabulusu, line 110: why this variable.. just call the function in next line

app/routers/sessions.py
```diff
@@ -15,13 +16,48 @@
...
     """converts string to datetime format"""
     return datetime.fromisoformat(datetime_str)
 
 
+def shuffle_question_order(question_sets):
+    question_order = []
+    bucket_size = Settings().subset_size
+
+    global_index = 0  # Track global index across all questions
+    logger.info(question_sets)
```
suryabulusu, line 32: what? why log all questions.

suryabulusu (commented):
looks ok to me.. needs tests

app/routers/sessions.py
Asp-irin: ohh missed it! will remove this.

app/routers/sessions.py
Asp-irin: yes but how about having checking if in the last_session question_order is present or not if not then we can add identity order here itself rather than updating the db.

app/routers/sessions.py
Asp-irin: yep! done

app/routers/sessions.py
Asp-irin: Done added a script for all the previous sessions to have the linear mapping if it doesn't have question_order in it!

suryabulusu (approved):
looks great...approved. will merge after frontend is fixed.

PR #123, opened 2025-04-14
fix: prevent multiple active sessions for the same quiz

### LINKED ISSUE #103: Multiple devices accessing quiz with same quiz_id user_id combo
Add checks to ensure that multiple devices do not use the same quiz at the same time.
- This creates issues in session logic.

Session route logic assumes that there is only one session running at a time wrt (user_id + quiz_id) combination

ananyajain10:
Fixes #103 
Problem: The create_session route allowed multiple sessions for the same user_id + quiz_id combination.
Solution: 
Added a pre-check in the create_session endpoint:
Checks for any existing session with has_quiz_ended: false.
If found, returns a 409 Conflict with an appropriate message.
This ensures that only one active session can exist per user_id + quiz_id at a time.

Show this if session is active and new session creation for same  user_id + quiz_id
image


suryabulusu:
Hey -- we allow for multiple sessions for the same user+quiz combination. Every time a user resumes a session, a new session is created. This allows us to have a sequence of session interactions.

What we don't want: multiple user+quiz combinations in parallel. 

So, you will have to reframe your logic. There's an open PR from my end for this. You can look at it. Please do reach out to me on discord for more info.

Also, you've changed way too many files. Something wrong with pre-commit? Please check.

Thanks!

PR #142, opened 2026-04-12
4 - feat: Local Redis Caching

deepansh96:
## What this PR does

During a 10K-user quiz test, the same quiz document gets read **~30K+ times per container** from MongoDB. Every student hitting the same quiz triggers the same database read over and over. This PR adds a **local Redis sidecar** to each ECS task that caches immutable data (quizzes, questions, organizations), turning those repeated MongoDB reads into sub-millisecond localhost cache hits.

The cache is purely additive — if Redis dies or isn't running, the app falls back to direct MongoDB reads with zero user impact. Caching is controlled by a `CACHE_ENABLED` environment variable that starts as `false` and gets flipped to `true` via Terraform after we verify everything works.

### 1. Cache infrastructure (`app/cache.py`, `app/settings.py`)

New `CacheSettings` class follows the same lazy accessor pattern as `MongoSettings` — never evaluated at import time. The cache module provides `cache_get()`, `cache_set()`, and a `cache_key()` helper for building namespaced keys like `cache:v1:quiz:{id}`.

Redis connects lazily on first use with automatic reconnect (5-second backoff). If Redis is unreachable, every cache operation silently returns `None` / no-ops — no exceptions ever bubble up to callers. Error logging is rate-limited to once per 60 seconds so a Redis outage doesn't spam the logs.

Wired into `app/main.py`'s lifespan as best-effort — Redis init happens after the MongoDB ping, and if it fails, the app starts normally.

### 2. Shared quiz loader (`app/cache.py`)

`get_cached_quiz(quiz_id)` is the single function all routes now use to read a quiz. It checks cache first, on miss reads from MongoDB, applies the backwards-compatibility fixup, caches the canonical result with a 1-hour TTL, and returns it. Returns `None` for not-found (routes handle their own 404s).

This replaces scattered `db.quizzes.find_one()` calls across quiz, form, and session routes — every route now serves the exact same canonical quiz document.

### 3. Extracted shared services

**`app/services/quiz_fixups.py`** — The backwards-compatibility fixup function (`update_quiz_for_backwards_compatibility`) was embedded in the quiz router. Moved it to a neutral module so the shared quiz loader can use it without importing from a router.

**`app/services/omr.py`** — The OMR option-count aggregation pipeline was duplicated in both `quizzes.py` and `forms.py` (nearly identical code). Extracted into a shared helper that also fixes a fragile design: the old code used a **positional array** where list index had to match the question set order. Now it's a **dictionary keyed by question_set_id** — if a required key is missing, it returns a clear 500 error instead of silently serving wrong data.

### 4. Route caching (all routers)

Every route that reads quiz/question/org data now checks the cache first:

- **Quiz and form routes** use `get_cached_quiz()`, then `deepcopy()` the result before applying route-specific transforms (answer hiding, OMR shaping, single-page formatting). The deepcopy prevents mutations from leaking into the cached data.
- **Questions route** caches individual questions (1h TTL) and paginated question lists with normalized cache keys (falsey skip → 0, falsey limit → "all"). Empty `[]` results are valid cache hits.
- **Organizations route** caches org-auth lookups with a shorter **5-minute TTL** since API keys can be revoked.
- **Session routes** use the cached quiz loader for quiz reads, but session documents and answers are **never cached** — they're mutable and always go directly to MongoDB.

### 5. Backfill migration script

`app/scripts/backfill_quiz_backwards_compatibility.py` is a one-time script that applies the backwards-compat fixup to all existing quiz documents in MongoDB. Run this **before** enabling the cache so the runtime fixup becomes a no-op and every quiz is already in canonical form when it gets cached.

Supports `--dry-run` to preview changes. Idempotent — running it twice is safe.

### 6. Terraform Redis sidecar

Both testing and prod ECS task definitions get a Redis 7.2.5-alpine sidecar container:
- `essential: false` — if Redis crashes, the app container keeps running
- 128MB hard memory limit, 64MB soft reservation
- `maxmemory 64mb` with `allkeys-lru` eviction
- Health check via `redis-cli ping`
- Logs go to the existing ECS log group (not a separate one)

Cache environment variables (`CACHE_ENABLED=false`, `REDIS_URL`, `REDIS_MAX_CONNECTIONS`, `CACHE_NAMESPACE`) added to the app container in both environments.

### 7. Integration tests

11 new tests in `app/tests/test_cache_integration.py` covering:
- Cache miss → cache hit for quiz, form, question, and org-auth reads
- Cross-route shared cache keys (quiz cached by one route is a hit for another)
- Legacy quiz fixup: stripped fields restored via cache, MongoDB also updated
- OMR keyed map: missing question_set_id returns 500
- Redis-disabled fallback: all endpoints work with `CACHE_ENABLED=false`
- Empty paginated results: `[]` is a valid cache hit

Tests use `unittest.skipUnless(_redis_available())` so they're skipped gracefully when Redis isn't running — existing CI (no Redis) continues to pass.

## Rollout plan

> ⚠️ **IMPORTANT:** Run `terraform apply` on both testing and prod environments BEFORE merging. The CI/CD pipeline only swaps the app container image — it does NOT deploy the Redis sidecar or cache env vars. Without `terraform apply`, the cache-aware code will deploy without Redis.

1. Run backfill migration script on production MongoDB (`--dry-run` first)
2. `terraform apply` — deploys Redis sidecar with `CACHE_ENABLED=false`
3. Verify Redis sidecar is healthy in ECS console
4. Flip `CACHE_ENABLED=true` in Terraform and apply
5. Monitor cache hit/miss telemetry in logs
6. Load test on testing environment

## Test plan

- [x] All 133 existing tests pass (cache disabled by default)
- [x] 11 new cache integration tests pass (144 total)
- [ ] Run backfill migration script with `--dry-run` on prod
- [ ] `terraform plan` on testing environment
- [ ] Deploy to testing with `CACHE_ENABLED=false`, verify Redis sidecar starts
- [ ] Flip `CACHE_ENABLED=true` on testing, verify cache hit/miss logs
- [ ] Load test on testing environment


app/scripts/backfill_quiz_backwards_compatibility.py
```diff
@@ -0,0 +1,128 @@
...
+            continue
+
+        if args.dry_run:
+            print(f"  [dry-run] would update quiz {quiz_id}")
+            updated += 1
+            continue
+
+        apply_fixup(quiz)
+        try:
+            result = db.quizzes.update_one({"_id": quiz_id}, {"$set": quiz})
```
suryabulusu, line 108: quiz["_id"] and quiz_id are the same.. and we use this recipe in other places... but its a bit risky line 

maybe we can try 

  await db.quizzes.replace_one({"_id": quiz_id}, quiz) ?

or like explicitly remove quiz["_id"] from the dict before updating

app/services/quiz_fixups.py
```diff
@@ -0,0 +1,55 @@
...
+                    "skipped": 0,
+                }  # default
+
+    if is_backwards_compatibile:
+        logger.info("Quiz is already backwards compatible")
+        return
+
+    logger.info("Starting update for backwards compatibility")
+    db = get_quiz_db()
+    update_result = await db.quizzes.update_one({"_id": quiz_id}, {"$set": quiz})
```
suryabulusu, line 46: quiz["_id"] and quiz_id are the same.. and we use this recipe in other places... but its a bit risky line 

maybe we can try 

  await db.quizzes.replace_one({"_id": quiz_id}, quiz) ?

or like explicitly remove quiz["_id"] from the dict before updating

suryabulusu (commented):
one comment to check for safety -- on quiz_fixups.py .. coz this kinda runs on every request

PR #158, opened 2026-07-01
Ingest new-CMS chapter tests: POST /quiz/from-cms

pritamps:
Part of the **lms-cms-tests** effort. This is the **quiz-backend** half: map an assembled chapter-test JSON from the new CMS (nex-gen-cms) into a native quiz and store it, so the LMS session creator can serve CMS tests as live quizzes — replacing the legacy `sessionCreator` Lambda glue.

Pairs with nex-gen-cms PR #151 (the CMS routes this consumes).

## What

- **`POST /quiz/from-cms`** `{test_id, curriculum_id, grade_id, quiz_type}` → fetches the CMS assembled JSON, maps it, stores it, returns `{id, source_id, warnings}`.
- **`services/cms_ingest.py`** — fetch (Bearer `CMS_SERVICE_TOKEN`) + mapper.

## Mapping (contract locked with the CMS owner)

- **Type map**: `mcq_single_answer` / `matrix_match` / `comprehension` → `single-choice`; `mcq_multiple_answer` → `multi-choice`; `numerical_answer` → `numerical-integer`/`-float`.
- **Answers**: CMS 1-based option numbers → 0-based indices.
- **Marks**: cascade problem → section → subject → test (lowest set wins) → `correct = pos_marks[0]`, `wrong = -neg_marks[0]`. One question set per CMS section. Comprehension inlines its paragraph.
- **Multi-choice partial**: the standard JEE-Adv ladder (+k for k correct selected). A single per-set list suffices — the grader already gates partial on *no wrong selected* (`_is_subset` in `scoring.py`) and keys on `num_correct_selected`, and exact-match handles full marks. Confirmed against the scorer, not assumed.
- **Numerical ranges** `[low, high]` aren't gradeable by the engine (scalar only) → stores the low bound + emits a warning.

## Other

- Extracted `_insert_quiz_with_questions` from `create_quiz`; both endpoints share it and the same `Quiz` model validation (no behaviour change to `POST /quiz`).
- New `QuizSource` enum stamps `source="nex-gen-cms"`, distinct from the legacy free-text `"cms"` (old CMS) in BigQuery analytics.

## Verification

- **7 unit tests** for the mapper (single/multi/numerical/range/comprehension/cascade/empty).
- **Full end-to-end** on Python 3.9 + local Mongo against the live CMS (staging DB): `POST /quiz/from-cms` for a real chapter test → `201` → the stored quiz fetches back as a native quiz (question in `quiz.questions`, `source="nex-gen-cms"`, answer correctly hidden on the student-facing GET).
- black + flake8 clean.

> Note: the pinned `fastapi==0.75` / `pydantic==1.9` only import on Python ≤3.10 (fine — CI/staging use 3.9).

### DIFF
--- a/app/services/cms_ingest.py
+++ b/app/services/cms_ingest.py
@@ -1,0 +1,322 @@
+"""
+CMS -> quiz mapping for the LMS session-creation flow.
+
+The new CMS (nex-gen-cms) owns test content and exposes an assembled-test JSON at
+GET /api/service/test?id=&curriculum_id=&grade_id= (bearer-auth). This module fetches
+that JSON and maps it into the quiz format this service stores (quiz.quizzes /
+quiz.questions), so Gurukul renders a CMS-sourced test like any other quiz.
+
+Contract (locked with the CMS owner — see task lms-cms-tests):
+- Assembled shape: {"test": Test, "problems": [Problem, ...]}. `test.type_params` carries
+  the structure (subjects -> sections -> problem refs) and marks at four levels
+  (test / subject / section / problem). `problems` is a flat list of fully-resolved
+  problems (text, options, answer, paragraph) joined to their refs by id.
+- Answers are 1-based option numbers; the quiz engine wants 0-based indices.
+- Marks cascade problem -> section -> subject -> test; the lowest level that sets marks
+  wins. `pos_marks[0]` -> correct, `neg_marks[0]` -> wrong (as a negative).
+- Multi-choice partial marking uses the standard JEE-Adv preset (+1 per correct option
+  selected, no wrong selected; full marks handled by exact-match in the grader). A single
+  uniform partial list works per set regardless of each question's correct-count, because
+  the grader awards partial only for strict subsets keyed on num_correct_selected.
+- numerical_answer: single value -> scalar; 2-entry [low, high] range is not gradeable by
+  the engine today (scalar only), so we store the low bound and surface a warning.
+"""
+
+from typing import Any, Dict, List, Optional, Tuple
+
+import requests
+
+from schemas import QuizSource
+from settings import Settings
+
+settings = Settings()
+
+# CMS problem subtype -> quiz-engine question type. numerical_answer resolves to
+# numerical-integer/float at map time based on the answer value.
+CHOICE_TYPE_MAP = {
+    "mcq_single_answer": "single-choice",
+    "mcq_multiple_answer": "multi-choice",
+    "matrix_match": "single-choice",  # single-answer; table baked into the question HTML
+    "comprehension": "single-choice",  # 1:1, paragraph self-carried on the problem
+}
+
+
+class CmsIngestError(Exception):
+    """Raised when the CMS assembled-test JSON cannot be fetched or is unusable."""
+
+
+def fetch_assembled_test(
+    test_id: int, curriculum_id: int, grade_id: int
+) -> Dict[str, Any]:
+    """Fetch the assembled-test JSON from the new CMS. Raises CmsIngestError on failure."""
+    if not settings.cms_service_endpoint or not settings.cms_service_token:
+        raise CmsIngestError(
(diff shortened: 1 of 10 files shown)


app/services/cms_ingest.py
```diff
@@ -0,0 +1,325 @@
...
+            if not questions:
+                continue
+
+            if optional_problems:
+                warnings.append(
+                    f"section '{section.get('name') or section_index}' has optional "
+                    "problems; mandatory_count limits are not applied in v1 (all included)"
+                )
+
+            correct, wrong = _cascade_marks(None, section, subject, type_params)
```
suryabulusu: something weird happening here -- not like what sessionCreator does. 

so usually -- this is what i do 

cms section and mongo question set mostly map one is to one

provided a section has the same question type throughout

but if cms section has questions of different types

i break down that section further. the idea is to ensure absolutely that a question set in mongo has questions of the same type.

now wrt marks -- i basically override the individual question marking scheme of that set -- with the scheme of the first question of that set. this is usually fine -- especially for advanced-- when you are certain all questions of a set are of the same type, and therefore of the same marking scheme. 

i feel all of this is not being accounted for here. 

i am open to change logic here of course. this isn't sacrosanct, and honestly hacky from my end. but it generally has worked

app/services/cms_ingest.py
```diff
@@ -0,0 +1,325 @@
...
+                    has_multi_choice = True
+                    max_options = max(max_options, len(question.get("options") or []))
+                questions.append(question)
+                if question.get("graded", True):
+                    num_graded_questions += 1
+
+            if not questions:
+                continue
+
+            if optional_problems:
```
suryabulusu: i dont think optional is being accounted for well here.

So a section like “attempt 5 of 10” becomes “attempt 10 of 10”, inflating max marks, skipped counts, and percentages. 

This should not be a warning-only path for live quizzes. The mapper should preserve CMS optional attempt limits in max_questions_allowed_to_attempt, splitting where needed, or fail ingestion for optional sections.

app/services/cms_ingest.py
```diff
@@ -0,0 +1,325 @@
...
+            warnings.append(
+                f"problem {problem_id}: numerical with no answer -> ungraded"
+            )
+            question["type"] = "numerical-integer"
+            question["graded"] = False
+            return question, warnings
+        if len(answers) >= 2 and str(answers[0]) != str(answers[1]):
+            warnings.append(
+                f"problem {problem_id}: numerical range [{answers[0]}, {answers[1]}] "
+                "is not gradeable as a range by the engine; stored the low bound only"
```
suryabulusu: i wonder if this is right.. because we have this notion of tolerance.

if answer is 5.5 .. and user answered 5.3 . we would grade it correct.. coz tolerance is 0.5 i think. 

range would be 5 to 6

perhaps we should use low + high / 2 as the answer

suryabulusu: correction: current tolerance is 0.05 .. not 0.5

i wonder if we should just incorporate the range concept instead of this tolerance. 

but that could be a later PR

suryabulusu:
One metadata/settings note after pulling main: the branch now preserves the newer main changes (`require_all_questions` on Quiz and `matrix_rows` in the trimmed question projection). Separately, for this CMS ingest path, quiz-level metadata is still a bit sparse: `grade` is currently left as `None` even though `grade_id` is in the request, and CMS duration/instructions are not mapped onto `time_limit`/quiz instructions. Not blocking the main mapping discussion, but worth making an explicit call so the created Mongo quiz is self-contained where possible.

suryabulusu (approved):
all comments resolved. ready to merge after tests on staging

pritamps:
Note: pushed f06b805 after approval — deploy config only (SAM templates + workflows now provision CMS_SERVICE_ENDPOINT/CMS_SERVICE_TOKEN on the Lambda; no app-code change). Staging had neither var, which blocked the staging e2e you asked for.

PR #161, opened 2026-07-09
Add PATCH /quiz/{id} to patch session-editable settings in place

pritamps:
## What
New field-scoped endpoint **`PATCH /quiz/{quiz_id}`** that `$set`s only the session-editable settings on an existing quiz doc, in place (same `_id`): `title`, `shuffle`, `show_scores`, `review_immediate`, and `metadata.session_end_time`.

## Why
Part of the CMS quiz-sessions Phase 2 work (task `lms-cms-session-edit-regen`, stream 1). The LMS session-edit flow now writes each field directly to its source of truth instead of firing the legacy SNS → etl-data-flow `patch_session` (which KeyErrored on CMS sessions' absent `meta_data.course`). The quiz-taking frontend reads display/scoring settings from the quiz doc, so the LMS calls this endpoint to keep it in sync — without rebuilding the quiz or its questions.

## Behaviour
- Only fields present in the body are changed; empty body is a `200` no-op (`{"updated": []}`).
- `review_immediate` = "show answers immediately after submission".
- **Null-safe on `metadata`**: a dotted `$set` raises when `metadata` is explicitly `null`, so the whole subdoc is written in that case; absent metadata uses dot-notation.
- `404` on a missing quiz, and on a quiz deleted between the existence check and the write.
- No auth — consistent with the rest of the quiz router (CORS-gated).

## Tests
`app/tests/test_quizzes.py`: multi-field update, single-field isolation, no-op, null-metadata regression, 404.

## Deploy note
The af_lms edit route (separate PR) depends on this endpoint — **this must deploy first**, or CMS session edits will 502.

### DIFF
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -1,3 +1,5 @@
+from typing import Optional
+
 from fastapi import APIRouter, status, HTTPException, Query
 from fastapi.responses import JSONResponse
 from fastapi.encoders import jsonable_encoder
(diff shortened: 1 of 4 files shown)


app/routers/quizzes.py
```diff
@@ -224,6 +226,78 @@ async def create_quiz_from_cms(request: CmsQuizIngestRequest):
...
+
+    set_ops = {}
+    updated = []
+    for field in ("title", "shuffle", "show_scores", "review_immediate"):
+        value = getattr(request, field)
+        if value is not None:
+            set_ops[field] = value
+            updated.append(field)
+    if request.session_end_time is not None:
+        updated.append("session_end_time")
```
suryabulusu, line 493: seems to directly store the supplied session_end_time. 

But sessionCreator does not store raw session end time; it stores end_time + quiz.time_limit.max

Example: if a quiz window ends at 1:45 PM and the quiz duration is 3h, sessionCreator makes answers visible around 4:45 PM, not 1:45 PM

we do this because -- a person can start test around 1:30PM too. and can continue till 4:30 pm. had we shown answers to other students at 1:45 PM -- there's scope of copying. 

of course, this isn't perfect. some could pause and resume even after 4:45PM. but this was a slightly better choice.

now that i think of -- it should be called answer_visibility_time or something instead of session_end_time, but this requires some backfilling etc. for now we need not bother about it.

app/routers/quizzes.py
```diff
@@ -224,6 +226,78 @@ async def create_quiz_from_cms(request: CmsQuizIngestRequest):
     )
 
 
+class QuizPatchRequest(BaseModel):
```
suryabulusu, line 451: there are more things that can get patched

meta data test format
metadata grade
metadata single page header text

suryabulusu (commented):
left comments

app/routers/quizzes.py
pritamps: @suryabulusu Where can I look for this `end_time + quiz.time_limit.max` in code? Can't seem to find it anywhere

pritamps:
Thanks @suryabulusu — you're right on both counts, confirmed against the current `sessionCreator`. Pushed an update to this branch.

**On `session_end_time`:** agreed, the stored value needs to be the answer-visibility moment = window end **+ `time_limit.max`**, not the raw window end — matching `_quiz_answer_visibility_end_time` in `SessionCreator.py`. I've ported that logic into a new `services/quiz_time.py`: `PATCH` now reads `time_limit` off the existing quiz doc and stores `window_end + duration`, so the LMS just sends the raw window end and quiz-backend owns the offset (same place legacy computes it). The parser accepts ISO, 24h, and the legacy 12h `%I:%M:%S %p` format so the offset can't be silently dropped, and it logs a warning on anything unparseable. Leaving the `session_end_time` → `answer_visibility_time` rename for later as you suggested — not worth the backfill right now.

While wiring this I also found the **from-cms create** path never set `session_end_time` at all, so CMS-created quizzes had no answer-visibility gate (a `review_immediate=false` session would show answers immediately). Fixed that too — create now sets it with the same offset.

**On the other patchable fields (test_format / grade / single-page header):** rather than widen the field-scoped `PATCH` (those aren't editable in the LMS session-edit form — it only touches name/timing/show_scores/show_answers/shuffle/format), I've added a **regenerate-in-place** endpoint, `PUT /quiz/{id}/from-cms`. It re-pulls the corrected test and patches the existing quiz in place — same quiz/set/question `_id`s so the session and submitted attempts stay linked — refreshing content + those content-metadata fields, while preserving the session-edit settings. It mirrors the legacy regenerate path (`patch_quiz_in_mongo` Mode B).

One hardening beyond legacy: before any write it validates the regenerated test lines up positionally **and** that each position is the same problem (anchored on `source_id`) — a reorder or delete+add is refused with a 409 rather than silently remapping answer keys onto the wrong questions. One known gap: the in-place writes aren't wrapped in a transaction (same as the create path), so a mid-loop infra failure could leave a half-applied state — tracked to follow up with a replica-set transaction.

That makes this a complete quiz-backend PR (patch + regenerate); re-scoring existing attempts on regenerate will be the follow-up.

suryabulusu (approved):
approved -- but for this to work well, we need to make changes to LMS too

PR #162, opened 2026-07-10
Speed up ECS deployments

deepansh96:
## What changed

- deploy the testing ECS service for pull requests from branches in this repository
- keep merge-to-main deployments gated on a successful CI workflow
- cap ECS stability polling at 15 seconds
- replace the fixed smoke-test delay with retries
- cancel stale testing deployments and serialize production deployments
- reduce ALB and container health-check intervals from 30 seconds to 5 seconds
- reduce health-check timeouts from 5 seconds to 2 seconds
- reduce ALB target deregistration delay from 300 seconds to 5 seconds
- build ARM64 images on native GitHub ARM runners instead of QEMU
- persist BuildKit cache layers in each environment's ECR repository

Fork pull requests are intentionally skipped because they cannot safely receive AWS credentials.

## Why

Before this PR, successful testing deployments averaged 6m 04s end to end. The ECS step averaged 4m 07s, Docker builds averaged 1m 04s with 3-minute cold spikes, and the smoke test averaged 17s.

## Final measurement

A fresh dummy commit produced a successful final run:

- full workflow: **2m 44s**, down from **6m 04s** (**55% faster**)
- ECS deployment step: **1m 55s**, down from **4m 07s** (**53% faster**)
- Docker build and push: **8s**, down from a **1m 04s** average (**88% faster**)
- smoke test: **3s**, down from **17s** (**82% faster**)

Final run: https://github.com/avantifellows/quiz-backend/actions/runs/29096419424

A prior optimized run completed in 2m 12s. Repeated runs place the new steady range around 2-3 minutes; ECS/Fargate rollout timing is now the main source of variance.

Native ARM cache measurements:

- first build while seeding ECR cache: **27s**, versus **179-194s** historical cold builds
- subsequent cached builds: **8s**, with all Docker layers confirmed as cache hits

The testing ALB and task definition were updated before measurement. Production infrastructure should be applied only after this PR is reviewed and merged.

## Validation

- `actionlint` on both ECS workflows
- `terraform fmt -check` on the changed Terraform files
- `terraform validate` for testing and production
- `git diff --check`
- CI and repeated ECS testing deployments

### DIFF
--- a/.github/workflows/deploy_ecs_prod.yml
+++ b/.github/workflows/deploy_ecs_prod.yml
@@ -7,10 +7,6 @@ on:
     types: [completed]
     branches: [release]
 
-  # Temporary: for testing on the feature branch. Remove after first successful deploy.
-  push:
-    branches: [docs/migration-lambda-to-ecs]
-
 env:
   AWS_REGION: ap-south-1
   ECR_REPOSITORY: quiz-backend-prod
@@ -27,18 +23,14 @@ jobs:
   deploy:
     name: Build & Deploy
     runs-on: ubuntu-24.04-arm
-    # For workflow_run: only deploy if CI succeeded
-    # For push: always run
-    if: >-
-      github.event_name == 'push' ||
-      (github.event_name == 'workflow_run' && github.event.workflow_run.conclusion == 'success')
+    # Only deploy if CI succeeded
+    if: github.event.workflow_run.conclusion == 'success'
 
     steps:
       - name: Checkout
         uses: actions/checkout@v4
         with:
-          # workflow_run uses the head SHA from the CI run; push uses the current SHA
-          ref: ${{ github.event_name == 'workflow_run' && github.event.workflow_run.head_sha || github.sha }}
+          ref: ${{ github.event.workflow_run.head_sha }}
 
       - name: Configure AWS credentials
         uses: aws-actions/configure-aws-credentials@v4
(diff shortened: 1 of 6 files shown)


deepansh96:
@codex review

.github/workflows/deploy_ecs_prod.yml
suryabulusu: since we are cleaning up.. can we remove this

terraform/prod/alb.tf
```diff
@@ -21,15 +21,17 @@ resource "aws_lb_target_group" "quiz_backend" {
   vpc_id      = data.aws_vpc.default.id
   target_type = "ip" # Required for Fargate
 
+  deregistration_delay = 5
```
suryabulusu: This value is the load balancer drain time for old ECS tasks during deploys.

With 5s, deploys are faster, but any inflight prod request that takes more than 5s can be interrupted when the old task is removed.

example ones i can think of: ingest... or maybe session search and fetch, etc. 

i think -- for prod, we can increase it. maybe to 20s or 30s. 

we anyway dont deploy prod that frequently. 

it is still unlikely to have requests exactly during the time we deploy. but i think this is a safe change

suryabulusu (commented):
left two comments

.github/workflows/deploy_ecs_prod.yml
deepansh96: Removed the temporary feature-branch push trigger and simplified the workflow_run-only condition and checkout in 48d25a6.

terraform/prod/alb.tf
deepansh96: Agreed. Increased the production deregistration delay to 30 seconds while leaving testing at 5 seconds in 48d25a6.

suryabulusu (approved):
all changes addressed. approving

PR #164, opened 2026-07-16
cms ingest: set question metadata.chapter from assembled payload

pritamps:
## Problem

nex-gen-cms questions were landing in the quiz Mongo `questions` collection with `metadata.chapter = null`. `_problem_metadata` set `chapter_id` but never the chapter **name**. The downstream quiz-results ETL groups question responses by the chapter **name** (`groupby([..., "chapter"])`), so **every nex-gen-cms `chapter_test` produced empty chapter-level analytics** (and, before an ETL empty-frame guard, crashed the worker).

Scale: 0 of ~26k nex-gen-cms question-level rows had a chapter; ~68 sessions affected and growing.

## Fix

The assembled `/api/service/test` payload already carries chapter per problem — **verified live against staging** (`test 1700`: `chapter_id` + `chapter_name` on all 25/25 problems), and at the db-service source (`resource/test/{id}/problems`). So no nex-gen-cms change is needed.

- Add `_chapter_name()` to flatten the multilingual `chapter_name` list (`[{"chapter": "Mole Concept", "lang_code": "en"}]`) to the English name — mirrors the existing `_test_name()` helper.
- Set `metadata.chapter` from it in `_problem_metadata`. `chapter_id` was already read.

## Testing

- `pytest app/tests/test_cms_ingest.py` → 18 passed, including two new cases:
  - `test_chapter_name_mapped_from_assembled_payload` — picks the English name out of a multi-lang list; `chapter_id` set.
  - `test_missing_chapter_is_none` — no `chapter_name` → `chapter`/`chapter_id` stay `None`, no crash.
- Verified the real ingest contract end-to-end: staging `/api/service/test` returns chapter for every problem.

## Context

Complements the ETL-side fallback in **etl-data-flow** ([PR #116](https://github.com/avantifellows/etl-data-flow/pull/116)), which resolves chapter from AF DB when `metadata.chapter` is missing. This PR fixes the **source** so new quizzes carry chapter natively; once deployed, the ETL fallback becomes belt-and-suspenders and can retire. Tracked as DEBT D25.

Only affects **new** quizzes generated from CMS; already-created sessions rely on the ETL fallback + a re-sync.

### DIFF
--- a/app/services/cms_ingest.py
+++ b/app/services/cms_ingest.py
@@ -155,12 +155,29 @@ def _solutions(meta_data: Dict[str, Any]) -> List[str]:
     return out
 
 
+def _chapter_name(problem: Dict[str, Any]) -> Optional[str]:
+    """The assembled problem's `chapter_name` is a multilingual list
+    ([{"chapter": .., "lang_code": ..}]); return the English chapter name.
+
+    The downstream quiz-results ETL groups question responses by this chapter
+    name to build chapter-level analytics, so it must be a plain string.
+    """
+    entries = problem.get("chapter_name") or []
+    for entry in entries:
+        if entry.get("lang_code") == "en" and entry.get("chapter"):
+            return entry["chapter"]
+    return (entries[0].get("chapter") or None) if entries else None
+
+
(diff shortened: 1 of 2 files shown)


suryabulusu (approved):
perfect

PR #174, opened 2026-08-06
fix(cms-ingest): make curriculum_id/grade_id optional on /quiz/from-cms

suryabulusu:
## Problem

nex-gen-cms **#170** shortened its test URLs to `/test?id=<id>`, dropping `curriculum_id`/`grade_id`. `/quiz/from-cms` required all three, so a caller holding only a pasted CMS test link couldn't satisfy the contract:

- **quiz-creator** rejects the URL — *"Please provide a valid legacy or new CMS test URL"* (`isValidCmsUrl` demands all three)
- **sessionCreator** is worse: `parse_new_cms_link` returns `None` when any is missing and `process_row` gates on `if new_cms_ids:` — so the link **silently falls through** to another dispatch branch and the quiz is never built as a CMS test

## Why they aren't needed

A test is identified by its **`id` alone**. Per db-service **#651**: *"every problem can have exactly one `resource_curriculum` row, hence filtering is not required on `curriculum_id`"* — so curriculum never selected content.

Confirmed against prod for test 8901, which is tagged to two curriculum-grade pairs (`[{2,4},{1,1}]`):

| Params | problems | sha(problem ids) | marks |
|---|---|---|---|
| `curriculum_id=2&grade_id=4` (real pair) | 180 | `5ef3b3f67679` | 720 |
| `curriculum_id=1&grade_id=1` (its other real pair) | 180 | `5ef3b3f67679` | 720 |
| `curriculum_id=999&grade_id=999` | 180 | same | 720 |

Byte-identical. Multi-curriculum tagging is about where a test is *listed*, not which problems it contains.

## Change

- `CmsQuizIngestRequest.curriculum_id` / `.grade_id` → `Optional[int] = None`
- `fetch_assembled_test` builds its query params from what it was given: `id` always, curriculum/grade only when the caller knows them

Callers that know the real pair keep sending it. **af_lms is unchanged** — it derives both from its own exam-track/grade pickers (`curriculumIdForExamTrack` / `resolveGradeId`) and fails closed if grade won't resolve.

Applies to both `POST /quiz/from-cms` and `PUT /quiz/{id}/from-cms` (shared model + fetch).

## Dependency — satisfied

Until today the CMS 502'd on `/api/service/test?id=<id>` alone (`getTestProblems` forwarded the empty param into db-service, which did `String.to_integer("")`). db-service **#677** fixed that and is deployed to `release`; re-verified against prod just now — `?id=8901` alone returns **200 with all 180 problems**. So this PR needs no placeholder or workaround.

## Tests

3 new cases in `test_cms_ingest.py`: explicit pair passes through; both omitted → only `id` is sent; partial → only what was given is sent. 26 passed; `black` clean.

## Follow-ups

- **etl-data-flow** `fix/cms-short-link-parse` — sessionCreator accepts an id-only link (pushed)
- **quiz-creator** `fix/accept-short-cms-test-url` — relax `isValidCmsUrl` (ready)
- **nex-gen-cms** — `getTestProblems` calls `http.Error` without `return`, then nil-derefs in `fillProblemSubjects`; worth guarding independently

### DIFF
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -192,13 +192,8 @@ class CmsQuizIngestRequest(BaseModel):
     """
 
     test_id: int
-    # Optional: the CMS assembled-test endpoint ignores their VALUES — a test is identified
-    # by `id` alone and `/api/service/test` returns the same problems whichever
-    # curriculum/grade is passed (a test can be tagged to several curriculum-grade pairs).
-    # They are still sent downstream because that endpoint 502s when they are absent, so
-    # fetch_assembled_test substitutes a placeholder. Callers that know the real pair should
-    # keep passing it; callers holding only a test link — nex-gen-cms shortened its test URLs
-    # to `/test?id=<id>` — can now omit them.
+    # Optional: the CMS identifies a test by `id` alone and ignores these values. Callers
+    # holding only a shortened CMS link (`/test?id=<id>`) can omit them.
     curriculum_id: Optional[int] = None
     grade_id: Optional[int] = None
     quiz_type: str = QuizType.assessment.value
--- a/app/routers/quizzes.py
+++ b/app/routers/quizzes.py
@@ -192,8 +192,15 @@ class CmsQuizIngestRequest(BaseModel):
     """
 
     test_id: int
-    curriculum_id: int
-    grade_id: int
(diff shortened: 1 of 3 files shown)


suryabulusu:
merging as this is a small fix

